PR #1, opened 2015-06-17
add TimeInMilliseconds method

talwai:


statsd/statsd.go
```diff
@@ -185,6 +185,13 @@ func (c *Client) Set(name string, value string, tags []string, rate float64) err
 	return c.send(name, stat, tags, rate)
 }
 
+// Sends timing information in milliseconds. Flushed by statsd with percentiles, mean and other info
```
jmoiron: golint will complain about this because Go documentation for functions/methods start off like `<FunctionName> ...`

talwai: Fixed

jmoiron:
lgtm :shipit:

remh:
it's interesting to notice that it's the same exact thing as a histogram.

we actually treat |h and |ms the same exact way in the server 
https://github.com/DataDog/dd-agent/blob/master/aggregator.py#L739-L740

I know it's the one type used by etsy statsd but do we really need this ?

talwai:
Admitting that `|ms` is essentially an alias of `|h`, my inclination is to keep this change so that sinks modeled after the [go-metrics StatsdSink](https://github.com/armon/go-metrics/blob/master/statsd.go#L56-L59) can use this client to send an identical set of metric types. I think maintaining parity in the packets themselves is important so as not to mess with any dowstream tooling

PR #3, opened 2015-08-28
Fix UDP Dial for go 1.5

w-vi:
Tests were failing when using go 1.5. Datagrams were never processed by the server. The only real change is the usage of `net.DialUDP` instead of `net.Dial`.


remh:
Thanks a lot for this we'll review this shortly.

w-vi:
Any update, we would appreciate if this gets merged. Cheers

remh:
Sorry for the delay @w-vi this looks good to me.

Thanks a lot!

PR #6, opened 2015-12-04
Create LICENSE

aviau:
MIT requires that you include the license:

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.


truthbk:
@aviau Thank you very much for the heads-up! As the copyright information is not quite correct in the license docs we have decided to just include it ourselves from scratch. I'm not sure if there could be any legal implications modifying a set license copyright, so better safe than sorry!

PR #11, opened 2016-03-11
Split messages in maximum UDP message size when buffering - Refactor

truthbk:
A small PR bringing the great work from @sschepens in #7. Because we had taken a long time to review his PR I decided to just consolidate it here a little bit, and rewrite one of the functions and test in an effort to make the code a little more readable.

### DIFF
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -25,6 +25,7 @@ package statsd
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"math/rand"
 	"net"
@@ -33,6 +34,25 @@ import (
 	"sync"
 	"time"
 )
+
+/*
+OptimalPayloadSize defines the optimal payload size for a UDP datagram, 1432 bytes
+is optimal for regular networks with an MTU of 1500 so datagrams don't get
+fragmented. It's generally recommended not to fragment UDP datagrams as losing
+a single fragment will cause the entire datagram to be lost.
+
+This can be increased if your network has a greater MTU or you don't mind UDP
+datagrams getting fragmented. The practical limit is MaxUDPPayloadSize
+*/
+const OptimalPayloadSize = 1432
+
+/*
+MaxUDPPayloadSize defines the maximum payload size for a UDP datagram.
+Its value comes from the calculation: 65535 bytes Max UDP datagram size -
+8byte UDP header - 60byte max IP headers
+any number greater than that will see frames being cut out.
+*/
+const MaxUDPPayloadSize = 65467
 
 // A Client is a handle for sending udp messages to dogstatsd.  It is safe to
 // use one Client from multiple goroutines simultaneously.
@@ -46,6 +66,7 @@ type Client struct {
 	bufferLength int
 	flushTime    time.Duration
 	commands     []string
+	buffer       bytes.Buffer
 	stop         bool
 	sync.Mutex
 }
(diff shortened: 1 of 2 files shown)


statsd/statsd.go
```diff
@@ -121,30 +135,100 @@ func (c *Client) watch() {
...
+			needed = sepLength + len(cmd)
+		}
+
+		if buffer.Len()+needed <= maxSize {
+			if elem != 0 {
+				buffer.WriteString(sep)
+			}
+			buffer.WriteString(cmd)
+			elem++
+		} else if elem == 0 {
```
sschepens: This case does not cover every possibility, if I first send a small cmd and then a large one it would fall through to `else`.
Maybe it could be `else if needed > maxSize {`

sschepens: We also should check that UDP packets do not get larger than 65467 bytes, beyond that is where metrics begin to get cut, but UDP packet is never dropped.
Also, if we decide to flush large commands regardless of size, maybe we should remove check on `sendMsg`

statsd/statsd.go
```diff
@@ -121,30 +135,100 @@ func (c *Client) watch() {
...
+	}
 	return err
 }
 
 func (c *Client) sendMsg(msg string) error {
 	// if this client is buffered, then we'll just append this
 	if c.bufferLength > 0 {
+		// return an error if message is bigger than MaxPayloadSize
+		if len(msg) > MaxPayloadSize {
+			return errors.New("message size exceeds MaxPayloadSize, consider increasing it")
```
sschepens: This error message is incorrect now, `MaxPayloadSize` can't be increased.

statsd/statsd.go
```diff
@@ -121,30 +135,100 @@ func (c *Client) watch() {
...
-	data := strings.Join(c.commands, "\n")
-	_, err := c.conn.Write([]byte(data))
+	frames, flushable := joinMaxSize(c.commands, "\n", MaxPayloadSize)
+	var err error
+	cmdsFlushed := 0
+	for i, data := range frames {
+		_, e := c.conn.Write([]byte(data))
+		if e != nil {
+			err = e
+			for j := 0; j < i; j++ {
```
sschepens: This loop could be removed, by making variable `i` accesible outside of loop, and using this same loop instead of `for _, flushed := range flushable`

statsd/statsd.go
```diff
@@ -121,30 +135,100 @@ func (c *Client) watch() {
...
+				cmdsFlushed += flushable[j]
+			}
+			break
+		}
+	}
+	if err == nil {
+		for _, flushed := range flushable {
+			cmdsFlushed += flushed
```
sschepens: This loop could be removed, see previous comment

statsd/statsd.go
```diff
@@ -121,30 +135,100 @@ func (c *Client) watch() {
...
+		}
+	}
+
 	// clear the slice with a slice op, doesn't realloc
-	c.commands = c.commands[:0]
+	if cmdsFlushed == len(c.commands) {
+		c.commands = c.commands[:0]
+	} else {
+		//this case would cause an eventual realloc in the future...
+		c.commands = c.commands[cmdsFlushed:]
```
sschepens: What would happen if `commands` is full and we can't send any command on a `flush` call? This would lead to a sort of overflow, commands would grow larger than specified capacity.

statsd/statsd.go
sschepens: This:
```go
	for i, data := range frames {
		_, e := c.conn.Write([]byte(data))
		if e != nil {
			err = e
			break
		}
		cmdsFlushed += flushable[i]
	}
```
Should remove the use of both loops commented

sschepens:
I love that you're using `bytes.Buffer`, we could use this opportunity to further optimize things. We're returning a slice of strings that then get casted to a `[]byte`, and this is unnecessary, `joinMaxSize` could return a slice of `[]byte` that could directly be sent to connection.
Also, we could convert `separator` to []byte once at the start of `joinMaxSize` and just write it to buffer, this would also save further allocations from converting `string` to `[]byte`

statsd/statsd.go
```diff
@@ -121,30 +135,100 @@ func (c *Client) watch() {
...
 			return err
 		}
 	}
-	c.Unlock()
 	return nil
 }
 
+func joinMaxSize(cmds []string, sep string, maxSize int) ([]string, []int) {
+	var buffer bytes.Buffer
+	buffer.Reset() //just to be sure...
```
sschepens: This is unnecessary, golang [bytes doc](https://golang.org/pkg/bytes/#Buffer) specifically says:
> The zero value for Buffer is an empty buffer ready to use.

statsd/statsd.go
```diff
@@ -121,30 +135,100 @@ func (c *Client) watch() {
...
+			//if a single command is too big, flush it and hope the UDP fragmented packet
+			//doesn't get dropped.
+			frames = append(frames, cmd)
+			ncmds = append(ncmds, 1)
+		} else {
+			frames = append(frames, buffer.String())
+			ncmds = append(ncmds, elem)
+			buffer.Reset()
+			buffer.WriteString(cmd)
+			elem = 1
```
sschepens: Shouldn't this be `elem = 0`? this case only happens when a new frame must be created and so we should clean it to 0.

statsd/statsd.go
```diff
@@ -121,30 +135,100 @@ func (c *Client) watch() {
...
+		}
+	}
+
+	//add whatever is left!
+	frames = append(frames, buffer.String())
```
sschepens: We should only do this two instructions if buffer is not empty, if not we would be adding and empty frame.

sschepens:
@truthbk Sorry I got carried away with too many comments, i've opened a [12](https://github.com/DataDog/datadog-go/pull/12) with most of the changes I suggested. Please have a look.
Nice work!

PS: I'm still somehow worried about not clearing `c.commands` on errors.

truthbk:
@sschepens I've decided we should drop the offending command to avoid the "leak". Probably if we get to that point where the socket is failing to send the buffer, we have either worse problems or the command will continue to fail in future iterations, so let's just drop it.

Let's get this merged, let me know how you feel about it ;)

Thanks!

sschepens:
@truthbk seems good! Do you consider we should stop writing to socket as soon as we find one error and just let it retry the rest on next flush? (next flush is gonna happen sooner though because commands is not empty)
I'm ok with both behaviours

truthbk:
@sschepens that's a great question. I feel like once we get an error on the socket, it's likely we get another one immediately if we continue to write to it... so my guess is it's probably best to back-off. But this isn't set in stone :)

truthbk:
Gave this a whirl with the buffered client, and it behaved well with no apparent regressions. Tests are good. Original authorship by @sschepens respected in squashed commit ;)

Shall we merge?

sschepens:
> Shall we merge?

@truthbk yes please!

Thanks for preserving authorship, though you deserve much credit too!
Cheers! :beers:

PR #12, opened 2016-03-12
Split messages in maximum UDP message size when buffering - Refactor 2

sschepens:
Another small PR joining the work from @truthbk and refactoring it more to optimize it.

### DIFF
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -34,7 +34,7 @@ import (
 	"time"
 )
 
-const MaxPayloadSize = 65507
+const MaxPayloadSize = 1432
 
 // A Client is a handle for sending udp messages to dogstatsd.  It is safe to
 // use one Client from multiple goroutines simultaneously.
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -46,7 +46,7 @@ datagrams getting fragmented. The theoretical limit is 65467
 (65535 bytes Max UDP datagram size - 8byte UDP header - 60byte max IP headers)
 any number greater than that will see data loss.
 */
-var MaxPayloadSize = 1432
+const MaxPayloadSize = 1432
 
 // A Client is a handle for sending udp messages to dogstatsd.  It is safe to
 // use one Client from multiple goroutines simultaneously.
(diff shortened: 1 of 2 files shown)


statsd/statsd.go
```diff
@@ -120,36 +141,103 @@ func (c *Client) watch() {
...
+
+		if c.buffer.Len()+needed <= maxSize {
+			if elem != 0 {
+				c.buffer.Write(sepBytes)
+			}
+			c.buffer.WriteString(cmd)
+			elem++
+		} else {
+			tmpBuf := make([]byte, c.buffer.Len())
+			copy(tmpBuf, c.buffer.Bytes())
```
truthbk: Any reason why you don't want to use the String() method for the buffer? I guess efficiency, but we're ultimately doing something similar here with an allocation and a copy.

statsd/statsd.go
```diff
@@ -120,36 +141,103 @@ func (c *Client) watch() {
...
+	frames, flushable := c.joinMaxSize(c.commands, "\n", OptimalPayloadSize)
+	var err error
+	cmdsFlushed := 0
+	for i, data := range frames {
+		_, e := c.conn.Write(data)
+		if e != nil {
+			err = e
+			break
+		}
+		cmdsFlushed += flushable[i]
```
truthbk: much cleaner, thanks...

statsd/statsd.go
```diff
@@ -120,36 +141,103 @@ func (c *Client) watch() {
...
 }
 
 func (c *Client) sendMsg(msg string) error {
 	// if this client is buffered, then we'll just append this
+	c.Lock()
+	defer c.Unlock()
 	if c.bufferLength > 0 {
+		// return an error if message is bigger than OptimalPayloadSize
+		if len(msg) > MaxUDPPayloadSize {
+			return errors.New("message size exceeds MaxUDPPayloadSize")
```
truthbk: this is probably the right thing to do... We should include the size in the error message though - both the msg size and `MaxUDPPayloadSize`.

statsd/statsd.go
```diff
@@ -46,6 +66,7 @@ type Client struct {
 	bufferLength int
 	flushTime    time.Duration
 	commands     []string
+	buffer       bytes.Buffer
```
truthbk: :bulb:

statsd/statsd_test.go
```diff
@@ -198,6 +198,136 @@ func TestBufferedClient(t *testing.T) {
 
 }
 
+func TestJoinMaxSize(t *testing.T) {
```
truthbk: I think I fixed a few things in the test in the other refactor PR after you started your work. Would you mind bringing those in? Unfortunately I squashed those, so it might be a bit of a mess. Alternatively I could rebase to your changes there....

sschepens: I think i did bring in your latest changes, had some conflicts because you squashed, but your changes should be here

statsd/statsd.go
sschepens: Sure!

statsd/statsd.go
```diff
@@ -120,36 +141,103 @@ func (c *Client) watch() {
...
+	var ncmds []int
+	sepBytes := []byte(sep)
+
+	elem := 0
+	for _, cmd := range cmds {
+		needed := len(cmd)
+
+		//if a single command is too big, flush it and hope the UDP fragmented packet
+		//doesn't get dropped.
+		if needed > maxSize {
```
truthbk: I'm not sure about this. This could change the order of the messages. It's not really a big problem as such, and this implementation could possibly result in fewer messages being sent...

sschepens: I don't see this as a big issue, but what would you suggest?

truthbk: I think we can keep this as is, your approach optimizes the number of messages sent.

statsd/statsd.go
truthbk: Probably in a separate PR we could consider the possibility of truncating the associated message (which is the only way I can see a message being >64K).

statsd/statsd.go
truthbk: @sschepens could you take a look please? Let me know if you don't have the time, just trying to get this merged soon :)

sschepens: Sure, I see you pulled my changes to you pull request, do you want me to close this one?
You're right about commands ordering, maybe we can flush everything we had accumulated before flushing a big command, this would be less efficient, but would preserve order

truthbk: @sschepens hey buddy! Sorry I went MIA last week, plenty of work on other fronts. Do you want to make the changes to `joinMaxSize()`, so we preserve order? Or do I make them? Just want to make sure we don't both work on this...  Let me know!

sschepens: @truthbk Hey! I've been MIA as well, I've committed the changes now, could you check them and pull them to your branch?

truthbk:
Good stuff! Thanks a lot for that :+1:

truthbk:
Closing in favor of #11

PR #15, opened 2016-04-08
Add Incr, Decr, Timing and more docs.

gphat:
# Summary

Adds additional convenience methods `Incr`, `Decr` and an alias for `Timing` for `TimeInMilliseconds`. It also adds additional documentation for new and old functions.

# Motivation

Other Datadog clients (and StatsD in general) include an increment and decrement method for convenience. Perhaps I'm lazy. :)

The `TimeInMilliseconds` method feels prescriptive to me, especially since:
* From what I understand, golang — which I'm admittedly new to — tends to work in nanoseconds. [Duration](https://golang.org/pkg/time/#Duration) doesn't have a `Milliseconds()`, so it feels odd to pass nanoseconds to such a clearly named method.
* Datadog's timer support does not imply any specific unit. The percentiles and such that come out of a timer lack metadata and do not have a unit of `milliseconds` at all.

# Notes

This is a very subjective pull request and I'm new to golang so forgive me if this doesn't track with the intent or future plans for the library. I won't mind if this isn't a direction you want to take. Thanks for the consideration!

### DIFF
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -268,10 +268,25 @@ func (c *Client) Histogram(name string, value float64, tags []string, rate float
 	return c.send(name, stat, tags, rate)
 }
 
+// Decr is just Count of 1
+func (c *Client) Decr(name string, tags []string, rate float64) error {
+	return c.send(name, "-1|c", tags, rate)
+}
+
+// Incr is just Count of 1
+func (c *Client) Incr(name string, tags []string, rate float64) error {
+	return c.send(name, "1|c", tags, rate)
+}
+
 // Set counts the number of unique elements in a group.
 func (c *Client) Set(name string, value string, tags []string, rate float64) error {
 	stat := fmt.Sprintf("%s|s", value)
 	return c.send(name, stat, tags, rate)
+}
+
+// Timing sends timing information, it is an alias for TimeInMilliseconds
+func (c *Client) Timing(name string, value time.Duration, tags []string, rate float64) error {
+	return c.TimeInMilliseconds(name, value.Seconds()*1000, tags, rate)
 }
 
 // TimeInMilliseconds sends timing information in milliseconds.
--- a/statsd/README.md
+++ b/statsd/README.md
@@ -19,12 +19,19 @@ if err != nil {
 c.Namespace = "flubber."
 // Send the EC2 availability zone as a tag with every metric
 c.Tags = append(c.Tags, "us-east-1a")
-err = c.Gauge("request.duration", 1.2, nil, 1)
+
+// Do some metrics!
+err = c.Gauge("request.queue_depth", 12, nil, 1)
+err = c.Timing("request.duration", duration, nil, 1) // Uses a time.Duration!
+err = c.TimeInMilliseconds("request", 12, nil, 1)
+err = c.Incr("request.count_total", nil, 1)
+err = c.Decr("request.count_total", nil, 1)
+err = c.Count("request.count_total", 2, nil, 1)
 ```
 
 ## Buffering Client
 
-Dogstatsd accepts packets with multiple statsd payloads in them.  Using the BufferingClient via `NewBufferingClient` will buffer up commands and send them when the buffer is reached or after 100msec.
+DogStatsD accepts packets with multiple statsd payloads in them.  Using the BufferingClient via `NewBufferingClient` will buffer up commands and send them when the buffer is reached or after 100msec.
 
 ## Development
 
--- a/statsd/statsd_test.go
+++ b/statsd/statsd_test.go
@@ -9,6 +9,7 @@ import (
 	"reflect"
 	"strings"
 	"testing"
+	"time"
 )
 
 var dogstatsdTests = []struct {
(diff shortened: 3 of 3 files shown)


truthbk:
@gphat Thanks a lot! Good to see you contributing to the datadog-go repo as well! Always very much appreciated. Thanks a lot for the `Incr` and `Decr` helpers. I'll take a closer look at the `TimeInMilliseconds` implementation and see if it makes sense to make those changes.

Also, thanks a bunch for the docs! :heart_eyes:

gphat:
No problem @truthbk! Hope it helps and thanks for the kind words. New to golang, which OS X just autocorrected to "gland". Awesome. :smiling_imp:

statsd/statsd.go
```diff
@@ -268,12 +268,27 @@ func (c *Client) Histogram(name string, value float64, tags []string, rate float
...
+}
+
 // Set counts the number of unique elements in a group.
 func (c *Client) Set(name string, value string, tags []string, rate float64) error {
 	stat := fmt.Sprintf("%s|s", value)
 	return c.send(name, stat, tags, rate)
 }
 
+// Timing sends timing information, it is an alias for TimeInMilliseconds
+func (c *Client) Timing(name string, value float64, tags []string, rate float64) error {
```
paulmach: flyby:  I think it'd be more idiomatic go to pass in a time.Duration for the time. Usage would be something like
```
start := time.Now()
// do somethings
c.Timing("name", time.Since(start), nil, 1.0)
```

Then this function could be:
```
func (c *Client) Timing(name string, val time.Duration, tags []string, rate float64) error {
    return c.TimeInMilliseconds(name, float64(value/time.Millisecond), tags, rate)
}
```

truthbk: @gphat, I agree with @paulmach could you please tweak that minor issue?

gphat: Done!

The `Duration` is a good idea, but I still feel the usage of `TimeInMillseconds` is both not idiomatic — as evidenced by the need to convert this to millis — and confusing from an API perspective. Datadog's timers not being inherently millis but can be whatever time you like. It would save a cast and a division to make this nanos!

That being said, this does what I want now and I'm happy enough to have it merged. I of course defer to you as the maintainers!

statsd/statsd.go
```diff
@@ -285,8 +285,8 @@ func (c *Client) Set(name string, value string, tags []string, rate float64) err
 }
 
 // Timing sends timing information, it is an alias for TimeInMilliseconds
-func (c *Client) Timing(name string, value float64, tags []string, rate float64) error {
-	return c.TimeInMilliseconds(name, value, tags, rate)
+func (c *Client) Timing(name string, value time.Duration, tags []string, rate float64) error {
+	return c.TimeInMilliseconds(name, float64(value.Nanoseconds())/float64(time.Millisecond), tags, rate)
```
dsansotscea: I tend to use `value.Seconds()*1000` since `Seconds` already returns a `float64`. I think it looks a bit cleaner, but I have no idea if one way is faster or if it even makes a difference.

gphat: There's [a whole thread on this](https://groups.google.com/forum/#!topic/golang-nuts/pret7hjDc70) that I read when initially working on this patch. I'm content with whatever the library authors are content with.

gphat:
Ping?

harlow:
+1 to PR

Wish I had found this earlier today. Would be much nicer to be able to pass the `time.Duration`. I ended up having to do a little conversion in middleware:

```go
func metricsMiddleware(c *statsd.Client, next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		start := time.Now()
		c.Tags = append(c.Tags, "path:"+r.URL.Path)

		// serve next handler
		next.ServeHTTP(w, r)

		// track response time
		responseTime := convertToMilliseconds(time.Since(start))
		err := c.Gauge("request.response_time", responseTime, nil, 1)
		if err != nil {
			fmt.Printf("gauge error %v", err)
		}
	})
}

func convertToMilliseconds(d time.Duration) float64 {
	return float64(d.Nanoseconds() / int64(time.Millisecond))
}
```

statsd/statsd.go
truthbk: I think @dsansotscea is probably right it does read better, and is probably a little faster. That said this has been waiting for a long time, so if @gphat can't address this in a day or so, I'll merge and fix that up myself in a separate PR.

Thank you both!

truthbk:
@gphat so sorry for letting this sleep for so long. My bad totally for snoozing on this. Just one comment, it's really more of a readability thing than anything else, so if you don't have the spare cycles to address it, I'll merge this as is.

gphat:
No worries, @truthbk. I've pushed an amended commit to keep the history clean and am using the suggested change.

Let me know if I misunderstood!

varun06:
Question - Is this PR going to be merged soon? I am planning to use it and we have same convenience methods with normal statsd lib. If it is getting merged soon, then I will just use these :)

truthbk:
@varun06 merging right now!

truthbk:
@gphat thanks a ton for the change and the PR my man. Merging! 💯  :shipit:

gphat:
No problem!

PR #17, opened 2016-08-12
Jaime/svchk support

diasjorge:

PR #19, opened 2016-09-27
Export serviceCheckStatus so interfaces can be built for statsd.Client

Jasrags:
Currently it is not possible to build a full interface for statsd.Client due to serviceCheckStatus being unexported. Example:

```
type StatsdClient interface {
...
    SimpleServiceCheck(name string, status byte) error
...
}
```

Does not match the status.Client.SimpleServiceCheck signature. Exporting the type allows for the following:

```
type StatsdClient interface {
...
    SimpleServiceCheck(name string, status statsd.ServiceCheckStatus) error
...
}
```

Which fulfills the interface.

If you know of a different way to get this to work without changing the type please let me know.

### DIFF
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -325,7 +325,7 @@ func (c *Client) ServiceCheck(sc *ServiceCheck) error {
 }
 
 // SimpleServiceCheck sends an serviceCheck with the provided name and status.
-func (c *Client) SimpleServiceCheck(name string, status serviceCheckStatus) error {
+func (c *Client) SimpleServiceCheck(name string, status ServiceCheckStatus) error {
 	sc := NewServiceCheck(name, status)
 	return c.ServiceCheck(sc)
 }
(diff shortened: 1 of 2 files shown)


Jasrags:
I would go as far as to expose eventAlertType and eventPriority as well, but that it outside of this current PR.

Jasrags:
Ping.

Jasrags:
@truthbk Could you take a look at this please.

AlbinOS:
I'm interested by this PR as well.

Jasrags:
@jmoiron Could you take a look at this please.

truthbk:
@Jasrags sorry for the slowness... I was out for a few days and missed this. Will review this week.

truthbk:
This is all very straight-forward.... And we shouldn't break anything by exporting the type (the converse would be a lot more problematic). I can't really see how but I'm just trying to think if we might somehow break something for some users before merging.

Jasrags:
@truthbk That was my thought as well. It seems to be a fairly new introduction into the code with https://github.com/DataDog/datadog-go/pull/17/files

truthbk:
@Jasrags I don't see how this could raise any issues. Merging. Thanks for your contribution! :)

PR #21, opened 2016-11-02
Remove newlines from tags

sjung-stripe:
If someone puts a newline in a tag, the content after the newline will be parsed as a completely new statsd message (because newlines are considered to be a message separator in the protocol). One way or another we have to get rid of them. Since newline escapes are not supported in tags, we simply remove them altogether.

cc @ChimeraCoder @krisreeves-stripe


truthbk:
Nice catch @tummychow, sounds reasonable to remove them. Taking a look at the changes.... Thanks!

statsd/statsd_test.go
```diff
@@ -33,6 +33,7 @@ var dogstatsdTests = []struct {
 	{"", nil, "Set", "test.set", "uuid", []string{"tagA"}, 1.0, "test.set:uuid|s|#tagA"},
 	{"flubber.", nil, "Set", "test.set", "uuid", []string{"tagA"}, 1.0, "flubber.test.set:uuid|s|#tagA"},
 	{"", []string{"tagC"}, "Set", "test.set", "uuid", []string{"tagA"}, 1.0, "test.set:uuid|s|#tagC,tagA"},
+	{"", nil, "Count", "test.count", int64(1), []string{"hello\nworld"}, 1.0, "test.count:1|c|#helloworld"},
```
truthbk: Thank you for the tests! 👏  🙇

PR #22, opened 2016-11-16
Tighten the scope of sendMsg lock

akrylysov:
Right now `sendMsg` lock also affects non buffered client, but it needs to be used only in `append` function.


truthbk:
@akrylysov sorry for the slow response, I was out for a few days. This sounds reasonable. Thanks a bunch!

truthbk:
@akrylysov your PRs work is actually included in @theckman's work in #24, so I'm going to be merging that one. Thank you very much regardless. 🙇

PR #23, opened 2016-12-01
Experiment to improve the message building speed by using less Sprintfs

corsc:
This PR is not meant to be landed but rather to serve as the basis for a discussion.

We (Grab.com) use DD StatsD a lot and as such an integral and frequently used part of our code base, any performance improvement has great value.

Anyway, here are my questions:
* How do you feel about trading code-complexity for performance in this style?
* Would you be willing to accept changes from outside contributors? (I give all contributions freely and with no expectations)
* How far do you want to take this sort of refactoring?
* Do you have any style guidelines/preferences that would help contributions be accepted?

The results of running the benchmark

    BenchmarkFormatV1-8   	10000000	      3057 ns/op
    BenchmarkFormatV2-8   	20000000	      1769 ns/op
    BenchmarkFormatV3-8   	20000000	      1392 ns/op


irabinovitch:
@masci @truthbk any thoughts on this one?

masci:
Hi @corsc @irabinovitch I'm usually not the maintainer so this fell completely off of my radar, sorry for the late reply!

Regarding your questions:

 * I'm totally 👍  in killing the `Sprintf`s, we did the same in the APM agent with a sensible performance improvement. Regarding complexity, I personally find the API even more clean: `send` is called with the actual value and type assertions happen in only one place, this makes a lot of sense to me.
 * Yes, we're happy to accept contributions!
 * If you're afraid a refactoring could go too far, my suggestion is to open an issue so we can discuss the design in advance.
 * We don't have official guidelines, as long as a PR has a clear intent, is well described and the code tested, this is usually enough to consider a merge.

Thanks for your code, waiting for @truthbk to chime in, he knows way better than me the library!

truthbk:
Hi @corsc! Wow, I thought this was already being reviewed by someone else and it fell completely off my radar. I'm actually very in favor of this sort of thing. The code changes look real good to me, I'll test-drive to double-check nothing breaks, but I'd say this is a strong 👍  

I totally agree with @masci, I actually believe you cleaned up the interface! So it's a win-win.

We have a small conflict. Please let me know if you don't have the time to fix it, we've kept you waiting wayyy too long, I feel bad even asking you.

Thanks again for your work, and apologies for the delay.

corsc:
Sorry for leaving this idle on my part too.  I will get back to it soon

PR #25, opened 2016-12-08
Client.sendMsg(): check payload len for all messages; add tests

### LINKED ISSUE #24: bug: sendMsg function doesn't always validate length
Hey there,

I was doing some reading of the codebase and I think stumbled upon a bug in the `sendMsg` method. Specifically, it seems that the length of the message is only checked if the client is in a buffered mode:

* https://github.com/DataDog/datadog-go/blob/426ff2e0daab1153de70993f85773d91302d50a0/statsd/statsd.go#L224-L228

Looking at it further, it seems the comment above the length check is at odds with what the code actually codes:

* https://github.com/DataDog/datadog-go/blob/426ff2e0daab1153de70993f85773d91302d50a0/statsd/statsd.go#L225

It's hard to know what the expected behavior truly is because the tests aren't currently ran for the function because it's unexported. Even then it seems to be at adds with the code itself:

* https://github.com/DataDog/datadog-go/blob/426ff2e0daab1153de70993f85773d91302d50a0/statsd/statsd_test.go#L341-L351

If you export the function and try to run the test, it results in a panic which is probably why it is unexported:

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x58 pc=0x7ea3b]

goroutine 26 [running]:
panic(0x17df00, 0xc4200140c0)
        /usr/local/Cellar/go/1.7.4/libexec/src/runtime/panic.go:500 +0x1a1
testing.tRunner.func1(0xc42009c480)
        /usr/local/Cellar/go/1.7.4/libexec/src/testing/testing.go:579 +0x25d
panic(0x17df00, 0xc4200140c0)
        /usr/local/Cellar/go/1.7.4/libexec/src/runtime/panic.go:458 +0x243
github.com/DataDog/datadog-go/statsd.(*Client).flush(0xc4200ae2a0, 0x0, 0x0)
        /Users/theckman/go/src/github.com/DataDog/datadog-go/statsd/statsd.go:201 +0xdb
github.com/DataDog/datadog-go/statsd.(*Client).append(0xc4200ae2a0, 0xc4200a8c00, 0x598, 0xc4200a8c00, 0x598)
        /Users/theckman/go/src/github.com/DataDog/datadog-go/statsd/statsd.go:141 +0xab
github.com/DataDog/datadog-go/statsd.(*Client).sendMsg(0xc4200ae2a0, 0xc4200a8c00, 0x598, 0x0, 0x0)
        /Users/theckman/go/src/github.com/DataDog/datadog-go/statsd/statsd.go:229 +0x124
github.com/DataDog/datadog-go/statsd.TestSendMsg(0xc42009c480)
        /Users/theckman/go/src/github.com/DataDog/datadog-go/statsd/statsd_test.go:343 +0xee
testing.tRunner(0xc42009c480, 0x1be4e8)
        /usr/local/Cellar/go/1.7.4/libexec/src/testing/testing.go:610 +0x81
created by testing.(*T).Run
        /usr/local/Cellar/go/1.7.4/libexec/src/testing/testing.go:646 +0x2ec
exit status 2
```

In short, I think the `sendMsg` method should check the length of the message whether or not it's a buffered client. Secondly, I think either the function's logic needs to be updated or the comment/tests need to be updated to match the expected behavior. Lastly, I think the changes in #22 should also be applied to `sendMsg()` / `append()`.

theckman:
This change makes some enhancements to the `Client.sendMsg()` method.
Specifically the following changes have been made:

* check the length of the message, whether buffered or unbuffered, and
  return an error if it's too long
* take the `sync.Mutex` lock inside of the `Client.append()` method,
  where it's needed, instead of in `Client.sendMsg()`

In addition to the changes above, the unit tests of this method have
been enabled / refactored to actually work.

Fixes #24

Signed-off-by: Tim Heckman <t@heckman.io>

### DIFF
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -135,6 +135,8 @@ func (c *Client) watch() {
 }
 
 func (c *Client) append(cmd string) error {
+	c.Lock()
+	defer c.Unlock()
 	c.commands = append(c.commands, cmd)
 	// if we should flush, lets do it
 	if len(c.commands) == c.bufferLength {
@@ -218,16 +220,16 @@ func (c *Client) flush() error {
 }
 
 func (c *Client) sendMsg(msg string) error {
+	// return an error if message is bigger than MaxUDPPayloadSize
+	if len(msg) > MaxUDPPayloadSize {
+		return errors.New("message size exceeds MaxUDPPayloadSize")
+	}
+
 	// if this client is buffered, then we'll just append this
-	c.Lock()
-	defer c.Unlock()
 	if c.bufferLength > 0 {
-		// return an error if message is bigger than OptimalPayloadSize
-		if len(msg) > MaxUDPPayloadSize {
-			return errors.New("message size exceeds MaxUDPPayloadSize")
-		}
 		return c.append(msg)
 	}
+
 	_, err := c.conn.Write([]byte(msg))
 	return err
 }
--- a/statsd/statsd_test.go
+++ b/statsd/statsd_test.go
@@ -338,15 +338,94 @@ func TestJoinMaxSize(t *testing.T) {
 	}
 }
 
-func testSendMsg(t *testing.T) {
-	c := Client{bufferLength: 1}
-	err := c.sendMsg(strings.Repeat("x", OptimalPayloadSize))
+func TestSendMsg(t *testing.T) {
+	addr := "localhost:1201"
+	udpAddr, err := net.ResolveUDPAddr("udp", addr)
 	if err != nil {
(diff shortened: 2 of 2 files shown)


truthbk:
You're absolutely right @theckman, we should check for the message length for both types of clients, buffered and unbuffered. Also thank you and @akrylysov for moving the mutex to the right place. Definitely two clear improvements to the unbuffered client. 

Also, can't thank you enough for the new tests. Brilliant stuff! 🙇

PR #27, opened 2016-12-16
Add Counter method

rikonor:
Hi,

When trying to count events I realized that the methods `Count`, `Inc` and `Dec` actually get aggregated into a rate of occurrences.
So if `Count(1)` is called 10 times within a 10 second window, the actual metric being reported is 1 (signifying a rate of 1 occurrence per second).

I then found the following [DataDog/dd-agent/aggregator.py#L112](https://github.com/DataDog/dd-agent/blob/master/aggregator.py#L112) which implements a real `Count` metric.

Notice that the names here are switched, since `Count` was already taken by the `rate-of-occurrence` method.

### DIFF
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -258,6 +258,12 @@ func (c *Client) Count(name string, value int64, tags []string, rate float64) er
 	return c.send(name, stat, tags, rate)
 }
 
+// Counter tracks how many times something happened within
+func (c *Client) Counter(name string, value int64, tags []string, rate float64) error {
+	stat := fmt.Sprintf("%d|ct", value)
+	return c.send(name, stat, tags, rate)
+}
+
 // Histogram tracks the statistical distribution of a set of values.
 func (c *Client) Histogram(name string, value float64, tags []string, rate float64) error {
 	stat := fmt.Sprintf("%f|h", value)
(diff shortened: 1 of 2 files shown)


remh:
Hi @rikonor 
Thanks for this!

However you don't need it. You should use the `Count` method. It's true that it will compute the rate of occurrences per second, however in the frontend you can actually graph the actual number of events happening by using the `Calcule as Count where applicable option`.

Example:
image

I'm closing this issue.
Feel free to reach out to our support team if needed.

rikonor:
Hi @remh,

Thanks for the quick response!

Would you be able to elaborate a bit on the actual value that will be shown if using `Calculate as count where applicable`?

Is it an exact count or an approximation?

If it is an approximation, then I would argue that there's still cause to include the actual Counter metric.

remh:
If you call it 10 times during the 10 seconds bucket. The value shown will be 10.

It is an exact count, as the flush interval is always 10 seconds when you use dogstatsd.

rikonor:
Yes, but I am wondering how the value 10 is reached.
Is it by multiplying the rate-of-occurence by 10?
Technically that would work just as well.

Also, I am wondering - why have that [Count metric](https://github.com/DataDog/dd-agent/blob/master/aggregator.py#L112) if it is not exposed by the API?

Seems to me that the initial intention was to include it (as can be seen [here](https://github.com/DataDog/dd-agent/blob/master/aggregator.py#L883)) but then, someone missed adding it [here](https://github.com/DataDog/dd-agent/blob/master/aggregator.py#L747) as well. Unless I completely misread the code 😃

remh:
Yes, exactly, we multiply the rate of occurence by 10. As the agent does the opposite the value is exact.

The count metric is used by the collector part of the agent and is not needed in dogstatsd as we have the calcule as count where applicable option available now :)

skolodyazhnyy:
I'm trying to get real count for some of my metrics but it keep giving me rate instead :( I couldn't find  "Calcule as Count where applicable option" option. I also tried to suffix metric with `.as_count()` but it does not seem to work. Any ideas?

Also, as I understood count is `rate * time-frame`, is there way to show raw value?

remh:
@skolodyazhnyy can you open a support ticket? If you use the `Count` method the option `Calcule as Count where applicable` should be available.

https://www.datadoghq.com/support/

Thanks!

PR #31, opened 2017-03-24
client: add new constructor, accepting a connection

cihangir:
We want to persist metrics to a storage while/before sending it to datadog agent.


truthbk:
@cihangir apologies for not seeing this earlier, I was out on vacation. I'll be reviewing this shortly. Thanks for your patience.

wvdeutekom:
Had a chance to review this @truthbk ?

PR #32, opened 2017-05-04
Improve message building speed by using less fmt.Sprintf

corsc:
Resubmission of [PR23](https://github.com/DataDog/datadog-go/pull/23)

The results of running the benchmark
```
BenchmarkFormatV1-8   	10000000	      3057 ns/op
BenchmarkFormatV2-8   	20000000	      1769 ns/op
BenchmarkFormatV3-8   	20000000	      1392 ns/op
```

### DIFF
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -55,6 +55,19 @@ any number greater than that will see frames being cut out.
 */
 const MaxUDPPayloadSize = 65467
 
+/*
+Stat suffixes
+*/
+var (
+	gaugeSuffix     = []byte("|g")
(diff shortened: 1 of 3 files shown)


corsc:
Sorry, I tried to get fancy with git and not create a commit while rebasing and completely failed.

I am not sure if it should a separate ticket or not but `go test -v github.com/DataDog/datadog-go/statsd -run ^TestSendMsg$` does not work for me.  On my branch or the current master

truthbk:
@corsc I think we can create a separate ticket for the failing test, particularly since it's broken on `master` already.

PR #36, opened 2017-07-13
Fix and add a test for a race in the buffered client

ebroder:
Technically, assignments to booleans (and ints, and other
word-or-less-sized scalars) is atomic, but the Go race detector
considers it to be racy, meaning that closing a buffered client will
trigger the race detector.

Instead, notify the background goroutine via a channel, which isn't
racy. In the process, switch to using a Ticker so that we can stop
it (time.Tick will always leak).

Finally, add a test which would have caught this behavior (the
existing test for buffered clients does not, because it instantiates a
client without spawning the flush goroutine), and enable the race
detector in Travis so that this is detected going forward.


masci:
Hi @ebroder and thanks for your code. PR looks very good to me, paging @truthbk since he's the maintainer

cory-stripe:
Pinging @masci and/or @truthbk on this?

PR #37, opened 2017-08-07
Add UDS support

xvello:
Add Unix Domain Socket support, as implemented in DSD6.

Protocol and error handling follows the specs at https://github.com/DataDog/datadog-agent/wiki/Unix-Domain-Sockets-support

Documentation will be updated once code is OK.

### DIFF
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -29,7 +29,6 @@ import (
 	"fmt"
 	"io"
 	"math/rand"
-	"net"
 	"strconv"
 	"strings"
 	"sync"
@@ -62,12 +61,6 @@ traffic instead of UDP.
 const UnixAddressPrefix = "unix://"
 
 /*
-UDSTimeout holds the default timeout for UDS socket writes, as they can get
-blocking when the receiving buffer is full.
-*/
-const defaultUDSTimeout = 1 * time.Millisecond
-
-/*
 Stat suffixes
 */
 var (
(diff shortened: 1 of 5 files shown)


statsd/statsd.go
```diff
@@ -56,6 +56,18 @@ any number greater than that will see frames being cut out.
...
+UnixAddressPrefix holds the prefix to use to enable Unix Domain Socket
+traffic instead of UDP.
+*/
+const UnixAddressPrefix = "unix://"
+
+/*
+UDSTimeout holds the default timeout for UDS socket writes, as they can get
+blocking when the receiving buffer is full.
+*/
+const DefaultUDSTimeout = 1 * time.Millisecond
```
truthbk: Do we need to export this?

statsd/statsd.go
```diff
@@ -71,7 +83,14 @@ var (
 // A Client is a handle for sending udp messages to dogstatsd.  It is safe to
 // use one Client from multiple goroutines simultaneously.
 type Client struct {
+	// Address to send metrics to, needed to allow reconnection on error
+	addr net.Addr
+	// Established connection object, or nil if Connect needed
 	conn net.Conn
+	// function pointer to use to write data to the connection
+	write func([]byte) error
```
truthbk: 🍰  If we export `Write`, `Client` would implement the `Writer` interface, which might be desireable (I'm not sure just a thought). Although we might also want precisely the opposite to prevent the users from writing directly to the socket 🤔

statsd/statsd.go
```diff
@@ -149,6 +180,16 @@ func (c *Client) format(name string, value interface{}, suffix []byte, tags []st
 	return buf.String()
 }
 
+// SetWriteTimeout allows the user to set a custom UDS write timeout. Not supported for UDP
+func SetWriteTimeout(c *Client, d time.Duration) error {
+	if c.addr.Network() == "unixgram" {
+		c.writeTimeout = d
+		return nil
+	} else {
+		return errors.New("SetWriteTimeout: only supported for UDS connexions")
```
truthbk: tiny typo: `connexions` -> `connections`

statsd/statsd.go
truthbk: On a second thought, it would be more idiomatic to have this be an interface to `io.Writer`.

And have two different structs, one for UDP and one for UDS, both implementing `Writer`. We can assign the right one during the creation of the `Client`. Should be straight-forward.

statsd/statsd.go
```diff
@@ -68,14 +73,25 @@ var (
 	timingSuffix    = []byte("|ms")
 )
 
-// A Client is a handle for sending udp messages to dogstatsd.  It is safe to
+// A statsdWriter offers a standard interface regardless of the underlying
+// protocol. For now UDS and UPD writers are available.
+type statsdWriter interface {
+	Write(data []byte) error
```
cihangir: Hey @xvello and @truthbk 

Do you mind explaining why you chose a custom "Write" method and deviated from the std lib io.Writer interface?

I was fixing my [PR](https://github.com/DataDog/datadog-go/pull/31) and dismayed by this change.

If that was an oversight I can fix that in that PR as well.

masci: @cihangir we should make it implement `io.Writer` indeed, can you fix it in #31?

Issue #18, opened 2016-08-18
Buffered client not flushed on Close


dsansotscea:
If a client is buffered, the buffer is not flushed when `Close` is called, leading to potential data loss.

varun06:
It should be easy to do, just check and flush the buffer in client `Close`. I haven't looked at the code much though.

PR #40, opened 2017-09-12
Exported Flush() function of the client

colega:
We're using datadog to have metrics of panics, but sometimes nobody recovers those panics and as we're using a buffered client, we lose those events/metrics. 

This is slightly different from #18 as we don't want to close the client, maybe someone on the stack knows how to recover that panic, but we do want to make sure our event was sent, so the usage for this would be like:
```go
if err := recover(); err != nil {
   client.Event(...)
   client.Flush()
   panic(err)
}
```


Ackar:
I'm also interested in this. Is it possible to merge it?
Thanks!

PR #43, opened 2017-11-21
Preventing the (*udsWriter).Close method from panic 

jacek-adamek:
Under some circumstances, for example when an uds client has not written any data yet, the uds connection can be nil. In such cases when applications try to close statsd client it panics.

### DIFF
--- a/statsd/statsd_test.go
+++ b/statsd/statsd_test.go
@@ -146,6 +146,24 @@ func TestClientUDS(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


statsd/statsd_test.go
```diff
@@ -146,6 +146,24 @@ func TestClientUDS(t *testing.T) {
 	}
 }
 
+func TestClientUDSClose(t *testing.T) {
```
masci: 🍰

jacek-adamek:
Thanks!

PR #46, opened 2018-01-25
Make statsdWriter implement io.Writer

masci:
* As PR title, `statsdWriter` now implements `io.Writer`.
 * Removed a dependency from kernel settings that made tests fail on different platforms
 * For the same reasons, removed some error type checking from the UDS client that was broken on mac
 * Added 1.9 and removed 1.4 from the test matrix

### DIFF
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -77,7 +77,7 @@ var (
 // A statsdWriter offers a standard interface regardless of the underlying
 // protocol. For now UDS and UPD writers are available.
 type statsdWriter interface {
-	Write(data []byte) error
+	Write(data []byte) (n int, err error)
 	SetWriteTimeout(time.Duration) error
 	Close() error
 }
@@ -112,7 +112,7 @@ func New(addr string) (*Client, error) {
 		}
 		return NewWithWriter(w)
 	}
-	w, err := newUdpWriter(addr)
+	w, err := newUDPWriter(addr)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


arbll:
👍

statsd/statsd.go
```diff
@@ -77,7 +77,7 @@ var (
 // A statsdWriter offers a standard interface regardless of the underlying
 // protocol. For now UDS and UPD writers are available.
 type statsdWriter interface {
-	Write(data []byte) error
+	Write(data []byte) (n int, err error)
```
truthbk: 🍰 

This is a necessary fix for sure, the "problem" is it breaks the current implementation. All good to merge to master, we should probably tag a new major?

statsd/statsd_test.go
```diff
@@ -540,9 +540,9 @@ func TestSendMsgUDP(t *testing.T) {
 		t.Error("Expected error to be returned if message size is bigger than MaxUDPPayloadSize")
 	}
 
-	longMsg := strings.Repeat("x", MaxUDPPayloadSize)
+	message := "test message"
```
truthbk: Not sure if there was a legit reason to send messages at the `MaxUDPPayloadSize`? I believe they were maybe trying to test there was no unexpected fragmentation of the statsd payload unless `len(message) > MaxUDPPayloadSize`?

I'm really not sure, just food for thought.

statsd/statsd.go
masci: Agreed, moved the milestone `2.0` and we can release accordingly

statsd/statsd_test.go
masci: I suspect it was for that reason, problem is this test would depend on the kernel settings of the hosts running the tests, IMO a bit out of scope for a unit test so I just lowered the message to a size that's not supposed to cause fragmentation.

PR #53, opened 2018-03-04
Reduce `Client.format` allocations

vcabbage:
Hello, I was reviewing this library as I may start using it at work. I noticed that there are some opportunities for reducing allocations. This can be taken farther, but I figured I'd open the conversation by working on `Client.format`.

The stack allocated buffer is 200 bytes simply because I figured it would be plenty large. The size can be adjusted if someone has a more educated guess at a reasonable upper bound of the message size.

Thanks!

Summary:

* Avoid conversions to/from `string` as much as possible.
* Preallocate a sufficiently large `[]byte` in `Client.format` and use it in such a way that it doesn't escape. This allows the compiler to stack allocate the buffer.
* Copied test cases from existing tests to use in benchmark.

Results (`go version go1.10 darwin/amd64`):

```
name               old time/op    new time/op    delta
ClientFormat/0-8      347ns ± 1%     236ns ± 2%  -32.08%  (p=0.000 n=20+18)
ClientFormat/1-8      343ns ± 1%     236ns ± 2%  -31.31%  (p=0.000 n=17+20)
ClientFormat/2-8      450ns ± 1%     250ns ± 2%  -44.49%  (p=0.000 n=20+19)
ClientFormat/3-8      539ns ± 1%     261ns ± 2%  -51.63%  (p=0.000 n=18+20)
ClientFormat/4-8      450ns ± 1%     249ns ± 2%  -44.55%  (p=0.000 n=20+20)
ClientFormat/5-8      266ns ± 1%      79ns ± 2%  -70.12%  (p=0.000 n=19+19)
ClientFormat/6-8      275ns ± 1%      86ns ± 1%  -68.58%  (p=0.000 n=17+18)
ClientFormat/7-8      567ns ± 3%     355ns ± 2%  -37.38%  (p=0.000 n=17+20)
ClientFormat/8-8      570ns ± 2%     357ns ± 2%  -37.40%  (p=0.000 n=18+20)
ClientFormat/9-8      228ns ± 1%      74ns ± 1%  -67.45%  (p=0.000 n=17+19)
ClientFormat/10-8     237ns ± 1%      77ns ± 1%  -67.55%  (p=0.000 n=20+19)
ClientFormat/11-8     319ns ± 1%      86ns ± 3%  -72.94%  (p=0.000 n=20+19)
ClientFormat/12-8     348ns ± 1%     154ns ± 2%  -55.85%  (p=0.000 n=20+17)

name               old alloc/op   new alloc/op   delta
ClientFormat/0-8       152B ± 0%       32B ± 0%  -78.95%  (p=0.000 n=20+20)
ClientFormat/1-8       152B ± 0%       32B ± 0%  -78.95%  (p=0.000 n=20+20)
ClientFormat/2-8       168B ± 0%       32B ± 0%  -80.95%  (p=0.000 n=20+20)
ClientFormat/3-8       184B ± 0%       32B ± 0%  -82.61%  (p=0.000 n=20+20)
ClientFormat/4-8       168B ± 0%       32B ± 0%  -80.95%  (p=0.000 n=20+20)
ClientFormat/5-8       168B ± 0%       32B ± 0%  -80.95%  (p=0.000 n=20+20)
ClientFormat/6-8       168B ± 0%       32B ± 0%  -80.95%  (p=0.000 n=20+20)
ClientFormat/7-8       168B ± 0%       32B ± 0%  -80.95%  (p=0.000 n=20+20)
ClientFormat/8-8       184B ± 0%       48B ± 0%  -73.91%  (p=0.000 n=20+20)
ClientFormat/9-8       160B ± 0%       32B ± 0%  -80.00%  (p=0.000 n=20+20)
ClientFormat/10-8      160B ± 0%       32B ± 0%  -80.00%  (p=0.000 n=20+20)
ClientFormat/11-8      176B ± 0%       32B ± 0%  -81.82%  (p=0.000 n=20+20)
ClientFormat/12-8      200B ± 0%       32B ± 0%  -84.00%  (p=0.000 n=20+20)

name               old allocs/op  new allocs/op  delta
ClientFormat/0-8       3.00 ± 0%      1.00 ± 0%  -66.67%  (p=0.000 n=20+20)
ClientFormat/1-8       3.00 ± 0%      1.00 ± 0%  -66.67%  (p=0.000 n=20+20)
ClientFormat/2-8       4.00 ± 0%      1.00 ± 0%  -75.00%  (p=0.000 n=20+20)
ClientFormat/3-8       4.00 ± 0%      1.00 ± 0%  -75.00%  (p=0.000 n=20+20)
ClientFormat/4-8       4.00 ± 0%      1.00 ± 0%  -75.00%  (p=0.000 n=20+20)
ClientFormat/5-8       4.00 ± 0%      1.00 ± 0%  -75.00%  (p=0.000 n=20+20)
ClientFormat/6-8       4.00 ± 0%      1.00 ± 0%  -75.00%  (p=0.000 n=20+20)
ClientFormat/7-8       4.00 ± 0%      1.00 ± 0%  -75.00%  (p=0.000 n=20+20)
ClientFormat/8-8       4.00 ± 0%      1.00 ± 0%  -75.00%  (p=0.000 n=20+20)
ClientFormat/9-8       3.00 ± 0%      1.00 ± 0%  -66.67%  (p=0.000 n=20+20)
ClientFormat/10-8      3.00 ± 0%      1.00 ± 0%  -66.67%  (p=0.000 n=20+20)
ClientFormat/11-8      3.00 ± 0%      1.00 ± 0%  -66.67%  (p=0.000 n=20+20)
ClientFormat/12-8      6.00 ± 0%      1.00 ± 0%  -83.33%  (p=0.000 n=20+20)
```

### DIFF
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -33,6 +33,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"unicode/utf8"
 )
 
 /*
@@ -96,7 +97,7 @@ type Client struct {
 	// BufferLength is the length of the buffer in commands.
 	bufferLength int
 	flushTime    time.Duration
-	commands     []string
+	commands     [][]byte
 	buffer       bytes.Buffer
 	stop         chan struct{}
 	sync.Mutex
@@ -134,7 +135,7 @@ func NewBuffered(addr string, buflen int) (*Client, error) {
 		return nil, err
 	}
 	client.bufferLength = buflen
-	client.commands = make([]string, 0, buflen)
+	client.commands = make([][]byte, 0, buflen)
 	client.flushTime = time.Millisecond * 100
 	client.stop = make(chan struct{}, 1)
 	go client.watch()
@@ -143,37 +144,40 @@ func NewBuffered(addr string, buflen int) (*Client, error) {
 
 // format a message from its name, value, tags and rate.  Also adds global
 // namespace and tags.
-func (c *Client) format(name string, value interface{}, suffix []byte, tags []string, rate float64) string {
-	var buf bytes.Buffer
+func (c *Client) format(name string, value interface{}, suffix []byte, tags []string, rate float64) []byte {
+	// preallocated buffer, stack allocated as long as it doesn't escape
+	buf := make([]byte, 0, 200)
+
 	if c.Namespace != "" {
-		buf.WriteString(c.Namespace)
+		buf = append(buf, c.Namespace...)
 	}
-	buf.WriteString(name)
-	buf.WriteString(":")
+	buf = append(buf, name...)
+	buf = append(buf, ':')
 
 	switch val := value.(type) {
 	case float64:
-		buf.Write(strconv.AppendFloat([]byte{}, val, 'f', 6, 64))
+		buf = strconv.AppendFloat(buf, val, 'f', 6, 64)
 
 	case int64:
-		buf.Write(strconv.AppendInt([]byte{}, val, 10))
+		buf = strconv.AppendInt(buf, val, 10)
 
 	case string:
-		buf.WriteString(val)
+		buf = append(buf, val...)
 
 	default:
 		// do nothing
 	}
-	buf.Write(suffix)
+	buf = append(buf, suffix...)
 
 	if rate < 1 {
-		buf.WriteString(`|@`)
-		buf.WriteString(strconv.FormatFloat(rate, 'f', -1, 64))
+		buf = append(buf, "|@"...)
+		buf = strconv.AppendFloat(buf, rate, 'f', -1, 64)
 	}
 
-	writeTagString(&buf, c.Tags, tags)
+	buf = appendTagString(buf, c.Tags, tags)
 
-	return buf.String()
+	// non-zeroing copy to avoid referencing a larger than necessary underlying array
+	return append([]byte(nil), buf...)
 }
 
 // SetWriteTimeout allows the user to set a custom UDS write timeout. Not supported for UDP.
@@ -203,7 +207,7 @@ func (c *Client) watch() {
 	}
 }
 
-func (c *Client) append(cmd string) error {
+func (c *Client) append(cmd []byte) error {
 	c.Lock()
 	defer c.Unlock()
 	c.commands = append(c.commands, cmd)
@@ -216,7 +220,7 @@ func (c *Client) append(cmd string) error {
 	return nil
 }
 
-func (c *Client) joinMaxSize(cmds []string, sep string, maxSize int) ([][]byte, []int) {
+func (c *Client) joinMaxSize(cmds [][]byte, sep string, maxSize int) ([][]byte, []int) {
 	c.buffer.Reset() //clear buffer
 
 	var frames [][]byte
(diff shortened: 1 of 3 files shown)


vcabbage:
CI failed because I used [`B.Run()`](https://golang.org/pkg/testing/#B.Run), which was added in Go 1.7.

Since they're benchmarks and not used by consumers of the library or during CI, I added a build tag to omit them when compiled with < Go 1.7. I could also not use that API, but it does make the benchmarks nicer.

corsc:
Nice results 👍

zippolyte:
Hi @vcabbage, thanks for your interest !

While it is true this improves performance a bit, I don't feel like the trade off is worth it. 
You end up with a less flexible API, where you'd also need to add some checking mechanism to avoid panics caused by a buffer too small. Additionally, this makes the code a little less readable, not great for maintainability and contributing.
If there was no trade off to this improvement, I would definitely take it. Right now though, I am not really in favor of merging it.

gzussa:
Hi @vcabbage, This is great. Thanks for making this improvement :).
@ofek Looks good to me

statsd/statsd.go
```diff
@@ -143,37 +143,40 @@ func NewBuffered(addr string, buflen int) (*Client, error) {
...
 // format a message from its name, value, tags and rate.  Also adds global
 // namespace and tags.
-func (c *Client) format(name string, value interface{}, suffix []byte, tags []string, rate float64) string {
-	var buf bytes.Buffer
+func (c *Client) format(name string, value interface{}, suffix []byte, tags []string, rate float64) []byte {
+	// preallocated buffer, stack allocated as long as it doesn't escape
+	buf := make([]byte, 0, 200)
+
 	if c.Namespace != "" {
-		buf.WriteString(c.Namespace)
```
gbbr: Have you tried `strings.Builder`?

vcabbage: I didn't try it when originally working on this PR. Go 1.10 (where `strings.Builder` was introduced) had only been out for a month or so at the time.

Theoretically `strings.Builder` should have similar performance, possibly excepting some of the `strconv` calls. If it's more readable using that API my opinion is that it would be worth testing out.

gbbr: I would be curious to hear if `strings.Builder` introduces further improvement. I was surprised to see how big the difference is between it and `bytes.Buffer`. It was supposedly built exactly for this use case.

IMHO given how critical this path is, performance has priority over readability.

vcabbage: Having reviewed this code again, I'm doubtful that it would be an improvement but I'd be interested to know if my reasoning is incorrect. Part of the improvement is due to convincing the compiler to stack allocate the buffer by having a fixed initial size. Even with `var b strings.Builder; b.Grow(200)` I'm not sure the compiler would see through it and do something similar.

Ultimately the message is written to the network as a `[]byte`, so I'm not sure there's much to be gained by converting to a string, even with the optimization that `strings.Builder` provides.

gbbr: If the message is processed and used as a `[]byte`, then you're likely right. Although, [recent changes](https://github.com/golang/go/commit/ca0c449a6b1c6ecc75169f93cffa8a5630740030) in the Go compiler indicate the two to be interchangeable.

Honestly, the only way to tell is to implement and run the benchmarks again.

PR #60, opened 2018-05-29
STDOUT Writer to support Lambda-type use cases.

platinummonkey:
Added a STDOUT writer to support lambda - type use cases where metrics are collected from STDOUT or logs.

This is used by specifying the `stdout://` address.

This by default uses the lambda structure: `MONITORING|<dogstatsd format>\n` but can be overridden with for example: `stdout://foo|/|baz` to produce `foo|<dogstatsd format>|baz\n` for example. The prefix and suffix are optional and will default.

Also updated a couple docs typos.

### DIFF
--- a/statsd/README.md
+++ b/statsd/README.md
@@ -31,12 +31,12 @@ err = c.Count("request.count_total", 2, nil, 1)
 
 ## Buffering Client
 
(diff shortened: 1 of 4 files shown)


arbll:
This is outdated

PR #62, opened 2018-08-16
UDS: make writer able to deal with concurrent writes

LeoCavaille:
Each commit message has more info.
I added a failing test that highlighted the issue first before adding the code to fix it.

The CI already runs with the race detector enabled so it should be able to test the change properly.


nabiber:
We tried this change and it killed performance. I would revert since it makes this library usable.

PR #65, opened 2018-08-25
Don't ignore nil clients

Aceeri:
Not panicking is fine, but visibility into what the problem is would be nice.


asf-stripe:
Hrm, this broke some of our tests - they expected a nil error when sending to a nil statsd client (that was the only way we had to work around #88 so far); I think this would be fine if we could get a singleton error value (that our tests can assert for) rather than a newly-allocated error.

PR #74, opened 2018-12-11
uds: chage Mutex to RWMutex for fast already-connected path

KJTsanaktsidis:
PR #62 added in a `Mutex` around the Unix socket writer's `Write` method,
to protect the logic that conditionally connected the socket if there
was not one already. However, because of the lock placement, it
serialises all statsd writes globally, which is likely to be a
performance bottleneck for highly parallel systems.

This commit turns that Mutex into a `RWMutex`. In the case where there is
already a socket set, only a `RLock` is needed. In the case where the
socket needs to be connected, the a full `Lock` is performed to protect
the swapping of `w.conn`.

Note that this change, I believe, will make `SetWriteTimeout`
meaningless; multiple goroutines will call `conn.SetWriteDeadline` in
parallel, and the behaviour of that is to extend the deadline for future
_and all in-flight_ requests. If new calls to `SetWriteDeadline` keep
getting made, the deadline on an existing blocked call to `Write` will
be continuously extended. I'm not really sure how to fix this or whether
it's worth fixing though.

### DIFF
--- a/statsd/uds.go
+++ b/statsd/uds.go
@@ -20,7 +20,7 @@ type udsWriter struct {
 	conn net.Conn
 	// write timeout
 	writeTimeout time.Duration
-	sync.Mutex   // used to lock conn / writer can replace it
+	sync.RWMutex // used to lock conn / writer can replace it
 }
 
 // New returns a pointer to a new udsWriter given a socket file path as addr.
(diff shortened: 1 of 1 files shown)


truthbk:
Closing in favor of https://github.com/DataDog/datadog-go/pull/84, there was pretty big change introduced and I was unable to push the rebase to your branch @KJTsanaktsidis, but your changes are still good and therein contained. Thank you.

PR #79, opened 2019-02-06
Increase performance with UDS

hush-hush:


statsd/uds.go
```diff
@@ -29,11 +43,74 @@ func newUdsWriter(addr string) (*udsWriter, error) {
...
+
+	writer := &udsWriter{
+		addr:                       udsAddr,
+		conn:                       nil,
+		writeTimeout:               defaultUDSTimeout,
+		datagramQueue:              make(chan []byte, 8192),
+		stopChan:                   make(chan struct{}, 128),
+		datagramsDroppedOutputChan: make(chan struct{}, 128),
+		datagramsDroppedQueueChan:  make(chan struct{}, 128),
+		datagramsTotalChan:         make(chan struct{}, 128),
```
LeoCavaille: Curious why we are using a chan here rather than a mutex or atomic values as it's a buffered chan this would be blocking if the telemetry loop is stuck?

statsd/uds.go
```diff
@@ -29,11 +43,74 @@ func newUdsWriter(addr string) (*udsWriter, error) {
...
+				datagramsDroppedOutput := w.datagramsDroppedOutput
+				w.datagramsDroppedOutput = 0
+				datagramsDroppedQueue := w.datagramsDroppedQueue
+				w.datagramsDroppedQueue = 0
+				datagramsDroppedTotal := w.datagramsDroppedTotal
+				w.datagramsDroppedTotal = 0
+				datagramsTotal := w.datagramsTotal
+				w.datagramsTotal = 0
+				go func() {
+					w.telemetryClient.Count("dogstatsd.client.datagrams_dropped_output", datagramsDroppedOutput, []string{}, 10)
```
LeoCavaille: why are the sample rates set to 10 here?

statsd/uds.go
arbll: Honestly I just thought it was the most go-like solution. However after some benchmarking I was actually planning to replace those by Atomic values which will probably have less impact.

statsd/uds.go
arbll: Nice catch, wrong c/p probably.

PR #81, opened 2019-03-07
add a non-blocking implementation for UDS

arbll:
## What does this PR do

Use a non-blocking UDS client implementation by default. Another PR will follow this one to allow choosing between the old blocking behavior that allowed error checking and this one.

### Non-blocking UDS client

- Write to the UDS in a single thread. Use a channel to buffer datagrams to send. 
- You no longer get an error back if there was an issue writing the datagram (like UDP) but you get one if the channel buffer is full. I think this is a fair tradeoff but this is breaking the original guarantees. 

### Motivation

More details are available in the original POC https://github.com/DataDog/datadog-go/pull/79.

### Notes

The old blocking implementation is kept in case someone started relying on error checking.

Being able to configure the client to use the blocking implementation requires some configuration refactoring. It will be done in https://github.com/DataDog/datadog-go/pull/82.

You should probably review this PR commit by commit.


statsd/statsd.go
```diff
@@ -106,7 +106,7 @@ type Client struct {
 // "unix:///path/to/socket".
 func New(addr string) (*Client, error) {
 	if strings.HasPrefix(addr, UnixAddressPrefix) {
-		w, err := newUdsWriter(addr[len(UnixAddressPrefix)-1:])
+		w, err := newAsyncUdsWriter(addr[len(UnixAddressPrefix)-1:])
```
truthbk: What if the user wants the blocking writer? Also, please see my note below bout using the async writer as a blocking writer by changing the channel depth.

statsd/uds_async.go
```diff
@@ -0,0 +1,119 @@
...
+import (
+	"fmt"
+	"net"
+	"time"
+)
+
+// asyncUdsWriter is an internal class wrapping around management of UDS connection
+type asyncUdsWriter struct {
+	// Address to send metrics to, needed to allow reconnection on error
+	addr net.Addr
```
truthbk: I would probably put `addr`, `conn` and `writeTimeout` in a separate structure that both the async and blocking writer structs would include in their definitions. 

edit: see note regarding using the async writer as a blocking writer.

statsd/uds_async.go
```diff
@@ -0,0 +1,119 @@
...
+	udsAddr, err := net.ResolveUnixAddr("unixgram", addr)
+	if err != nil {
+		return nil, err
+	}
+
+	writer := &asyncUdsWriter{
+		addr:         udsAddr,
+		conn:         nil,
+		writeTimeout: defaultUDSTimeout,
+		// 8192 * 8KB = 65.5MB
```
truthbk: Let's make sure this is clearly documented - this will have an immediate impact in memory footprint on higher throughput clients.

statsd/uds_async.go
```diff
@@ -0,0 +1,119 @@
...
+func (w *asyncUdsWriter) ensureConnection() (net.Conn, error) {
+	// Check if we've already got a socket we can use
+	currentConn := w.conn
+
+	if currentConn != nil {
+		return currentConn, nil
+	}
+
+	// Looks like we might need to connect - try again with write locking.
+	if w.conn != nil {
```
truthbk: This seems redundant with the above (https://github.com/DataDog/datadog-go/pull/81/files#diff-95ff682af2f28c4955d9ba7de9d1ef8dR100), am I missing something?

statsd/uds_async.go
```diff
@@ -0,0 +1,119 @@
...
+func (w *asyncUdsWriter) Write(data []byte) (int, error) {
+	select {
+	case w.datagramQueue <- data:
+		return len(data), nil
+	default:
+		return 0, fmt.Errorf("uds datagram queue is full (the agent might not be able to keep up)")
+	}
+}
+
+func (w *asyncUdsWriter) write(data []byte) (int, error) {
```
truthbk: This function is concurrency unsafe, we should maybe add a note in case it ever occurs to us we want multiple writers for added performance.... Not sure we'd ever be able to do that with deadlines anyways, so feel free to ignore this comment.

statsd/uds_async.go
```diff
@@ -0,0 +1,119 @@
...
+func newAsyncUdsWriter(addr string) (*asyncUdsWriter, error) {
+	udsAddr, err := net.ResolveUnixAddr("unixgram", addr)
+	if err != nil {
+		return nil, err
+	}
+
+	writer := &asyncUdsWriter{
+		addr:         udsAddr,
+		conn:         nil,
+		writeTimeout: defaultUDSTimeout,
```
truthbk: Do we want to make this TO overridable? Currently there's no real clean interface to do that, and since these are not exported it wouldn't allow users to achieve that. Same goes for the queue depth (8192).

I believe they're both quite sensitive parameters that should be configurable.

statsd/uds_blocking.go
```diff
@@ -0,0 +1,65 @@
...
+	addr net.Addr
+	// Established connection object, or nil if not connected yet
+	conn net.Conn
+	// write timeout
+	writeTimeout time.Duration
+	sync.Mutex   // used to lock conn / writer can replace it
+}
+
+// New returns a pointer to a new blockingUdsWriter given a socket file path as addr.
+func newBlockingUdsWriter(addr string) (*blockingUdsWriter, error) {
```
truthbk: I'm not sure we need a specific implementation for a blocking writer, I believe in effect the async writer with a datagram queue depth of 1 is essentially the same thing. There might be some details we want to look at - channel blocking vs mutex locking, etc, but I think it'd behave exactly as expected, it would perhaps reduce the amount of duped code here by thinking about that.

truthbk:
OK, I see #82 addresses some of my observations here, feel free to ignore anything you've addressed there ;)

statsd/statsd.go
arbll: Dealing with this in #82

statsd/uds_async.go
arbll: Should be good now

statsd/uds_blocking.go
arbll: The issue is error checking. Unlike UDP, UDS will error out if the agent does not receive the payload. With the blocking implementation you would get alerted if your message did not reach the agent.

However, in my opinion it's pretty flawed especially if the client buffers. In this case you only get one error on the call that actually sends the payload.

I think the best path might be:
- Merge this and #82
- (Optional) Set the blocking implementation as default
- Release 2.x
- Remove the blocking implementation
- Release 3.0

WDYT ? 
cc @xvello you might have an opinion

statsd/uds_async.go
arbll: Added a comment

statsd/uds_async.go
arbll: Agreed. I am still missing a PR to add a bunch of docs before releasing this and #82

statsd/uds_async.go
arbll: It will be in #82

statsd/uds_async.go
arbll: Let's discuss https://github.com/DataDog/datadog-go/pull/81/files#r266391555 and come back to this if it still make sense

arbll:
Just noticed tests are failing. It seems travis was not working previously. 
Seems like I can't use testify because it only supports Go 1.7+....

PR #82, opened 2019-03-08
configuration at client creation

arbll:
## What does this PR do

This PR adds a generic way to deal with the client configuration. I decided to use function options here for the implementation. This is discussable and definitely has it's pro and cons. However I prefer it over it's alternative, using a configuration struct, since it allows for dumb default value logic while using a struct would add quite a bit of complexity dealing with Golang zero values. I'm happy to discuss if someone has an opinion here.

## Motivation

The list of options is growing and some new ones require to be known at the client creation. Creating new constructors like `NewBuffered` is not scalable. 

Allowing options to be mutable after the client creation isn't a very good practice IMO.

### DIFF
--- a/statsd/options.go
+++ b/statsd/options.go
@@ -37,7 +37,7 @@ type Options struct {
 	WriteTimeoutUDS time.Duration
 }
 
-func resolveOptions(options []Option) *Options {
+func resolveOptions(options []Option) (*Options, error) {
 	o := &Options{
 		Namespace:             DefaultNamespace,
 		Tags:                  DefaultTags,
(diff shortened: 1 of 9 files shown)


statsd/options.go
```diff
@@ -0,0 +1,109 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return o, nil
+}
+
+// Option is a client option. Can return an error if validation fails.
+type Option func(*Options) error
```
truthbk: I'm a fan 👌

statsd/options.go
```diff
@@ -0,0 +1,109 @@
...
+	// DefaultNamespace is the default value for the Namespace option
+	DefaultNamespace = ""
+	// DefaultTags is the default value for the Tags option
+	DefaultTags = []string{}
+	// DefaultBuffered is the default value for the Buffered option
+	DefaultBuffered = false
+	// DefaultMaxMessagesPerPayload is the default value for the MaxMessagesPerPayload option
+	DefaultMaxMessagesPerPayload = 16
+	// DefaultBlockingUDS is the default value for the BlockingUDS option
+	DefaultBlockingUDS = false
```
truthbk: Blocking UDS the default behavior earlier, so this would change the behavior for customer clients possibly unwillingly. I personally would love the gains, but maybe the buffered UDS sender should be an opt-in feature?

statsd/options.go
```diff
@@ -0,0 +1,109 @@
...
+// WithMaxMessagesPerPayload sets the MaxMessagesPerPayload option.
+func WithMaxMessagesPerPayload(maxMessagesPerPayload int) Option {
+	return func(o *Options) error {
+		o.MaxMessagesPerPayload = maxMessagesPerPayload
+		return nil
+	}
+}
+
+// BlockingUDS sets the BlockingUDS option.
+func BlockingUDS() Option {
```
truthbk: `WithBlockingUDS` for consistency, perhaps?

statsd/options.go
arbll: Let's keep blocking in 2.x

PR #84, opened 2019-04-08
uds: chage Mutex to RWMutex for fast already-connected path (rebase)

truthbk:
This PR rebases #74 and includes the original changes submitted by @KJTsanaktsidis, it's merely a rebase after the introduction of the async non-blocking writer with #81

Thank you for your contribution @KJTsanaktsidis

### DIFF
--- a/statsd/uds_blocking.go
+++ b/statsd/uds_blocking.go
@@ -1,0 +1,98 @@
+package statsd
+
+import (
+	"net"
+	"sync"
+	"time"
+)
+
+/*
+UDSTimeout holds the default timeout for UDS socket writes, as they can get
+blocking when the receiving buffer is full.
+*/
+const defaultUDSTimeout = 1 * time.Millisecond
+
+// udsWriter is an internal class wrapping around management of UDS connection
+type udsWriter struct {
+	// Address to send metrics to, needed to allow reconnection on error
+	addr net.Addr
+	// Established connection object, or nil if not connected yet
+	conn net.Conn
+	// write timeout
+	writeTimeout time.Duration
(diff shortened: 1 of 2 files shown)


statsd/uds_blocking.go
```diff
@@ -63,3 +59,34 @@ func (w *blockingUdsWriter) Close() error {
 	}
 	return nil
 }
+
+func (w *udsWriter) ensureConnection() (net.Conn, error) {
```
arbll:
```suggestion
func (w * blockingUdsWriter) ensureConnection() (net.Conn, error) {
```

statsd/uds_blocking.go
```diff
@@ -63,3 +59,34 @@ func (w *blockingUdsWriter) Close() error {
...
+
+	newConn, err := net.Dial(w.addr.Network(), w.addr.String())
+	if err != nil {
+		return nil, err
+	}
+	w.conn = newConn
+	return newConn, nil
+}
+
+func (w *udsWriter) unsetConnection() {
```
arbll:
```suggestion
func (w * blockingUdsWriter) unsetConnection() {
```

KJTsanaktsidis:
Nice - thanks for merging!

Quick question, with the inclusion of the non-blocking version, is there any compelling reason to use the blocking one with this mutex over the non-blocking one other than error handling?

If not, I'll probably switch our app over to use the non-blocking one

arbll:
Hi @KJTsanaktsidis 

No reason apart from error checking. We are keeping it as the default implementation in 2.x and plan to release a new major that will make the async version the default and only implementation.

PR #90, opened 2019-05-29
Use an error constant for "nil client" errors

asf-stripe:
This PR adjusts the error behavior for nil clients to be a bit more consistent:

* Returns an error that can be checked in tests when a `nil` client is used (from the freshly-allocated errors in #65)
* Adjusts the error message:
  * bring it in line with [go error conventions](https://github.com/golang/go/wiki/Errors): No starting capital letter
  * mention that it's the `statsd` client that is `nil`

As mentioned in #65, this is affecting us in stripe/veneur#722; hope you all agree that this change will work.


statsd/statsd_test.go
```diff
@@ -556,6 +556,31 @@ func TestNilSafe(t *testing.T) {
 	assertNotPanics(t, func() { c.SimpleServiceCheck("", Ok) })
 }
 
+func TestNilError(t *testing.T) {
+	var c *Client
+	tests := []func() error{
+		func() error { return c.SetWriteTimeout(0) },
```
arbll: nit: This list is missing a few cases like `Incr` and `Decr`. For consistency we should add them all

asf-stripe: sgtm, I'll add them. I'll also unify this test and the `assertNotPanics`-using ones above, since they should cover essentially the same set of things.

asf-stripe:
Updated! PTAL, @arbll (:

PR #92, opened 2019-09-10
Add a NoOp client

goodspark:
Can be useful for testing and mocking for downstream library users.

Resolves #88

PR #100, opened 2019-10-25
Let NoOpClient implement all public methods of Client

skaji:
Address https://github.com/DataDog/datadog-go/pull/92#issuecomment-546185482

Currently we cannot build the following code:
```go
package main

import (
	"os"

	"github.com/DataDog/datadog-go/statsd"
)

func main() {
	var c statsd.ClientInterface
	if addr := os.Getenv("DATADOG_STATSD_ADDR"); addr != "" {
		var err error
		c, err = statsd.New(addr)
		if err != nil {
			panic(err)
		}
	} else {
		c = &statsd.NoOpClient{}
	}
	defer c.Close()

	c.Incr("foo", nil, 1)
}
```
```
❯ go build main.go
# command-line-arguments
./main.go:20:9: c.Close undefined (type statsd.ClientInterface has no field or method Close)
``` 

I think this should work.
This PR lets NoOpClient implement all public methods of Client so that the above code works.

### DIFF
--- a/statsd/noop.go
+++ b/statsd/noop.go
@@ -71,6 +71,21 @@ func (n *NoOpClient) SimpleServiceCheck(name string, status ServiceCheckStatus) 
 	return nil
 }
 
+// Close does nothing and returns nil
+func (n *NoOpClient) Close() error {
+	return nil
+}
+
+// Flush does nothing and returns nil
+func (n *NoOpClient) Flush() error {
+	return nil
+}
+
+// SetWriteTimeout does nothing and returns nil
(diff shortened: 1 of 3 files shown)


skaji:
Not related to this PR but, is it better for statsd.New to return ClientInterface instead of Client itself?

skaji:
@arbll Thanks!

skaji:
@arbll  Could you create a new git tag that includes this change?
Thanks!

arbll:
@skaji v3.2.0 is out

PR #103, opened 2019-11-11
send total number of metrics,events,sc as telemetry

arbll:


arbll:
Will get included in https://github.com/DataDog/datadog-go/pull/118

PR #107, opened 2019-11-21
statsd: close the stop channel when closing a statsd client to avoid leaking.

knusbaum:
This patch fixes a resource leak that occurs when closing a statsd client.
The Close function sent an empty struct to the stop channel, but there were two
goroutines waiting on the stop channel. Since the Close function did not close
the stop channel, one of the two goroutines would not exit, and would instead be
leaked along with any memory resources the statsd client held.

This patch causes the Close function to close the stop channel, ensuring any
goroutines listening on that channel will know the client has been stopped.

### DIFF
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -497,29 +497,21 @@ func (c *Client) SimpleServiceCheck(name string, status ServiceCheckStatus) erro
 	return c.ServiceCheck(sc)
 }
 
-// shutdown shuts down the goroutines associated with the client and waits for them.
-// It returns true if shutdown was performed, and false if the client was already shut down.
-func (c *Client) shutdown() bool {
-	c.Lock()
-	defer c.Unlock()
-	select {
-	case <-c.stop:
-		return false
-	default:
-	}
-	close(c.stop)
-	c.wg.Wait()
-	return true
-}
-
 // Close the client connection.
 func (c *Client) Close() error {
 	if c == nil {
 		return ErrNoClient
 	}
-	if c.shutdown() {
-		c.Flush()
-		return c.sender.close()
+	c.Lock()
+	select {
+	case <-c.stop:
+		c.Unlock()
+		return nil
+	default:
 	}
-	return nil
+	c.Unlock()
+	close(c.stop)
+	c.wg.Wait()
+	c.Flush()
+	return c.sender.close()
 }
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -489,23 +489,12 @@ func (c *Client) SimpleServiceCheck(name string, status ServiceCheckStatus) erro
 	return c.ServiceCheck(sc)
 }
 
-func (c *Client) shutdown() {
-	c.Lock()
-	defer c.Unlock()
-	select {
-	case <-c.stop:
-		return
-	default:
-	}
-	close(c.stop)
-}
-
 // Close the client connection.
 func (c *Client) Close() error {
 	if c == nil {
 		return ErrNoClient
 	}
-	c.shutdown()
+	close(c.stop)
 	c.Flush()
 	return c.sender.close()
 }
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -489,15 +489,23 @@ func (c *Client) SimpleServiceCheck(name string, status ServiceCheckStatus) erro
 	return c.ServiceCheck(sc)
 }
 
+func (c *Client) shutdown() {
+	c.Lock()
+	defer c.Unlock()
+	select {
+	case <-c.stop:
+		return
+	default:
+	}
+	close(c.stop)
+}
+
 // Close the client connection.
 func (c *Client) Close() error {
 	if c == nil {
 		return ErrNoClient
 	}
-	select {
-	case c.stop <- struct{}{}:
-	default:
-	}
+	c.shutdown()
 	c.Flush()
 	return c.sender.close()
 }
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -503,15 +503,15 @@ func (c *Client) Close() error {
 		return ErrNoClient
 	}
 	c.Lock()
+	defer c.Unlock()
 	select {
 	case <-c.stop:
-		c.Unlock()
 		return nil
 	default:
 	}
-	c.Unlock()
 	close(c.stop)
 	c.wg.Wait()
-	c.Flush()
+	c.flushUnsafe()
+	c.sender.flush()
 	return c.sender.close()
 }
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -190,6 +190,7 @@ type Client struct {
 	buffer        *statsdBuffer
 	telemetryTags []string
 	stop          chan struct{}
+	wg            sync.WaitGroup
 	sync.Mutex
 }
 
@@ -272,9 +273,16 @@ func newWithWriter(w statsdWriter, o *Options, writerName string) (*Client, erro
 	c.sender = newSender(w, o.SenderQueueSize, c.bufferPool)
 	c.flushTime = o.BufferFlushInterval
 	c.stop = make(chan struct{}, 1)
-	go c.watch()
-	go c.telemetry()
-
+	c.wg.Add(1)
+	go func() {
+		defer c.wg.Done()
+		c.watch()
+	}()
+	c.wg.Add(1)
+	go func() {
+		defer c.wg.Done()
+		c.telemetry()
+	}()
 	return &c, nil
 }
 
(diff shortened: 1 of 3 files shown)


statsd/statsd.go
```diff
@@ -489,15 +489,23 @@ func (c *Client) SimpleServiceCheck(name string, status ServiceCheckStatus) erro
 	return c.ServiceCheck(sc)
 }
 
+func (c *Client) shutdown() {
+	c.Lock()
+	defer c.Unlock()
+	select {
+	case <-c.stop:
```
cgilmour: Is this needed?
It looks like it's guarding against `shutdown()` being called more than once.

statsd/statsd.go
```diff
@@ -489,15 +489,23 @@ func (c *Client) SimpleServiceCheck(name string, status ServiceCheckStatus) erro
...
 // Close the client connection.
 func (c *Client) Close() error {
 	if c == nil {
 		return ErrNoClient
 	}
-	select {
-	case c.stop <- struct{}{}:
-	default:
-	}
+	c.shutdown()
```
gbbr: I'd like to recommend a simpler approach. Where the `watch` and `telemetry` goroutines are started, add:
```
c.wg.Add(1)
go func() {
    defer c.wg.Done()
    c.watch()
}()
c.wg.Add(1)
go func() {
    defer c.wg.Done()
    c.telemetry()
}()
```
And in this function (`Close`) just do:
```
close(c.stop)
c.wg.Wait()
```

statsd/statsd.go
arbll: Closing a closed channel panics

statsd/statsd.go
```diff
@@ -489,15 +489,23 @@ func (c *Client) SimpleServiceCheck(name string, status ServiceCheckStatus) erro
 	return c.ServiceCheck(sc)
 }
 
+func (c *Client) shutdown() {
+	c.Lock()
```
arbll: I don't think we need to hold the client lock here

statsd/statsd.go
knusbaum: I added this initially to prevent a panic if `Close()` was called more than once.

On closer inspection, the current implementation already panics if `Close()` is called more than once, because it calls `c.sender.close()`, which unconditionally closes `c.sender`'s stop channel.

I'll refactor this.

statsd/statsd.go
knusbaum: This was here to prevent a race between two routines calling `Close()` at the same time. 

On closer inspection, I realize it's already illegal to call `Close()` more than once, so this is not necessary.

statsd/statsd.go
knusbaum: I don't know if we need the wait group, but I can add it. 

I wrote and used the `shutdown()` function rather than just `close(c.stop)` to prevent a panic if `Close()` was called more than once. 

I looked closer and realize that it will panic anyway if called more than once in `c.sender.close()`, so removing `shutdown()` and just calling `close(c.stop)` is sufficient.

knusbaum:
@gbbr I've put up the simplest fix. Let me know if you think I should add the wait group stuff.

statsd/statsd.go
gbbr: > On closer inspection, the current implementation already panics if `Close()`
In my opinion this is not a reason to continue that trend and I prefer your version. An instrumentation library should never crash a host application, regardless of user error. I'd make sure `Close` never panics.

statsd/statsd.go
gbbr: The definition of a graceful shutdown is to wait for all goroutines to exit. If you exit the process, or close the statsd client but don't wait for the goroutines to exit, that's also a leak.

Closing that channel doesn't guarantee that the goroutine exits and finishes before the `Close` call returns. Technically, it is definitely possible for a `Close` call to finish and the statsd client might not be closed because the `select` loop in any of those goroutines can always trigger the other case before the stop channel.

statsd/statsd.go
gbbr: If users can do something, it's not illegal. But I suspect @arbll is right and this doesn't need a lock, even if called in parallel. Channels are concurrent safe, what are we guarding against?

As a side note, if we wanna go completely leak-free on the close, we should also fix this bit: https://github.com/DataDog/datadog-go/blob/87e3273d8c230a3abfbbe9dc16b9aa935b3af496/statsd/sender.go#L114

It also closes, but doesn't wait for the goroutines to exit. Here, we can use a simpler mechanism, similar to what we have in our own tracer:
```
s.stop <- struct{}
<-s.stop
```
The first send on the stop channel will cause the `sendLoop` to exit, and the receive will be triggered once the goroutine completely if the channel closes in a defer around [this line](https://github.com/DataDog/datadog-go/blob/87e3273d8c230a3abfbbe9dc16b9aa935b3af496/statsd/sender.go#L90):
```
defer close(s.stop)
```

We use similar mechanisms in dd-trace-go

knusbaum:
@arbll Interested to hear your thoughts.

statsd/statsd.go
knusbaum: 👍 I'll fix the sender panic here as well.

The lock is necessary because multiple goroutines can go through the select's default branch at once.
There's no assurance that a goroutine that takes the default branch will call `close(c.stop)` before another goroutine gets through select's default branch, and so it's possible for more than one goroutine to call `close(c.stop)` unless the select and close are protected.

statsd/statsd.go
knusbaum: 👍 I'll add the waitgroup stuff to ensure the routines have shut down before returning.

knusbaum:
I updated this to fix a few issues.
`Close()` now properly shuts down and waits for the client's goroutines to exit.
`Close()` can now be called multiple times without a panic.
The `sender`'s `close()` method now waits for the sender's goroutine to exit.

statsd/statsd.go
gbbr: > The lock is necessary because multiple goroutines can go through the select's default branch at once.

Indeed. Very correct 👍

statsd/statsd.go
```diff
@@ -489,15 +497,29 @@ func (c *Client) SimpleServiceCheck(name string, status ServiceCheckStatus) erro
 	return c.ServiceCheck(sc)
 }
 
+// shutdown shuts down the goroutines associated with the client and waits for them.
+// It returns true if shutdown was performed, and false if the client was already shut down.
+func (c *Client) shutdown() bool {
```
gbbr: IMO this will be much easier to follow if we drop this method and just move this logic into `Close`. The bool return value is slightly unusual. I don't see why we'd need a separate method. I think a good reason to have a separate method can be:
* parent method is getting huge and code is hard to follow (not the case)
* code is re-used (also not the case)

I think none of the above is applied in this case.

statsd/statsd.go
```diff
@@ -489,15 +497,29 @@ func (c *Client) SimpleServiceCheck(name string, status ServiceCheckStatus) erro
...
+		return false
+	default:
+	}
+	close(c.stop)
+	c.wg.Wait()
+	return true
+}
+
 // Close the client connection.
 func (c *Client) Close() error {
```
gbbr: Extra bonus points for a simple test :)
```
func TestClosePanic(t *testing.T) {
     c := New("...")
     c.Close()
     c.Close()
}
```

statsd/statsd.go
```diff
@@ -494,10 +502,16 @@ func (c *Client) Close() error {
...
 	}
+	c.Lock()
 	select {
-	case c.stop <- struct{}{}:
+	case <-c.stop:
+		c.Unlock()
+		return nil
 	default:
 	}
+	c.Unlock()
```
gbbr: I am worried about this locking mechanism. It is still possible that one goroutine takes hold of the lock while another one takes the default branch? Should we maybe either keep it simple (no locks and risk a highly unlikely race where someone calls Close in parallel multiple times) or lock the entire call?

knusbaum: We can't lock the entire `Close` call, because `c.Flush()` needs to be able to obtain the lock. It *is* possible one goroutine takes control of the lock while another is past it, but that's OK. ~~The way the lock is used ensures two goroutines can't both close the channel.~~  Actually I need to move `c.Unlock()` after `close(c.stop)`. Then it will ensure two can't both close the channel. (oops)

This is why I had the `shutdown` function, so I could lock the part of this logic that needed to be locked and `defer c.Unlock()` to avoid the risky and more complicated manual unlocking. 

I don't think it's wise to introduce a race that causes a panic, no matter how unlikely, if we can avoid it.

knusbaum:
Still need to write the unit test for the race condition.

statsd/statsd_test.go
```diff
@@ -434,3 +434,12 @@ func TestEntityID(t *testing.T) {
 		t.Errorf("Expecting empty default tags, got %v", client.Tags)
 	}
 }
+
+func TestClosePanic(t *testing.T) {
+	c, err := statsd.New("localhost:8125")
+	if err != nil {
+		t.Fatal(err)
+	}
```
arbll: Nit: While legacy test are using the Go's raw test api, new tests should use testify
```go
assert.Nil(t, err)
```

bai: Sorry for jumping in uninvited, another option would also be:
```
assert.NoError(t, err)
```

gbbr: Nobody's uninvited <3

PR #108, opened 2019-11-22
shard buffers to reduce lock contention

arbll:
### What does this PR do?

We had a single buffer we wrote all serialized metrics to. It was creating contention when multiple Goroutines where pushing metrics at the same time. This PR tries to address this issue by instead using a configurable number of buffers to reduce the chance two Goroutines write to the same buffer at the same time.
To implement this, we need a way to decide to which buffer we should write when a metric is submitted.
Currently, this PR uses the metric name to do so:
https://github.com/DataDog/datadog-go/blob/12805b4c7a2528b930591bd6d14fea0e031d6aa2/statsd/statsd.go#L389-L390

### Benchmarks

Using `bench.RunParallel`. Each goroutine sends metrics with a different name:
```
benchmark                      old ns/op     new ns/op     delta
BenchmarkStatsdUDP-4          219           105           -52.05%
BenchmarkStatsdUDS-4          217           110           -49.31%
```

### Notes

- Sharding by metric name might not be optimal is some scenarios.
- Buffering might get a bit worse as some of the buffers might get rarely written to which would lead to datagrams containing few metrics.

### TODO

- [ ] add MIT licence & copyright to fnv1a || import it as a dependency

### DIFF
--- a/statsd/client_test.go
+++ b/statsd/client_test.go
@@ -145,16 +145,16 @@ func TestBufferedClient(t *testing.T) {
 
 	dur, _ := time.ParseDuration("123us")
 
-	client.Incr("ic", nil, 1)
-	client.Decr("dc", nil, 1)
-	client.Count("cc", 1, nil, 1)
-	client.Gauge("gg", 10, nil, 1)
-	client.Histogram("hh", 1, nil, 1)
-	client.Distribution("dd", 1, nil, 1)
-	client.Timing("tt", dur, nil, 1)
-	client.Set("ss", "ss", nil, 1)
+	client.Incr("ab", nil, 1)
+	client.Decr("ab", nil, 1)
+	client.Count("ab", 1, nil, 1)
+	client.Gauge("ab", 10, nil, 1)
+	client.Histogram("ab", 1, nil, 1)
+	client.Distribution("ab", 1, nil, 1)
+	client.Timing("ab", dur, nil, 1)
+	client.Set("ab", "ss", nil, 1)
 
-	client.Set("ss", "xx", nil, 1)
+	client.Set("ab", "xx", nil, 1)
 	client.Flush()
 	if err != nil {
 		t.Errorf("Error sending: %s", err)
@@ -169,15 +169,15 @@ func TestBufferedClient(t *testing.T) {
 	}
 
 	expected := []string{
-		`foo.ic:1|c|#dd:2`,
-		`foo.dc:-1|c|#dd:2`,
-		`foo.cc:1|c|#dd:2`,
-		`foo.gg:10|g|#dd:2`,
-		`foo.hh:1|h|#dd:2`,
-		`foo.dd:1|d|#dd:2`,
-		`foo.tt:0.123000|ms|#dd:2`,
-		`foo.ss:ss|s|#dd:2`,
-		`foo.ss:xx|s|#dd:2`,
+		`foo.ab:1|c|#dd:2`,
+		`foo.ab:-1|c|#dd:2`,
+		`foo.ab:1|c|#dd:2`,
+		`foo.ab:10|g|#dd:2`,
+		`foo.ab:1|h|#dd:2`,
+		`foo.ab:1|d|#dd:2`,
+		`foo.ab:0.123000|ms|#dd:2`,
+		`foo.ab:ss|s|#dd:2`,
+		`foo.ab:xx|s|#dd:2`,
 	}
 
 	for i, res := range strings.Split(result, "\n") {
--- a/statsd/fnv1a.go
+++ b/statsd/fnv1a.go
@@ -1,0 +1,39 @@
+package statsd
+
+const (
+	// FNV-1a
+	offset32 = uint32(2166136261)
+	prime32  = uint32(16777619)
+
+	// init32 is what 32 bits hash values should be initialized with.
+	init32 = offset32
+)
+
+// HashString32 returns the hash of s.
+func hashString32(s string) uint32 {
+	return addString32(init32, s)
+}
+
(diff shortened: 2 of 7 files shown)


statsd/fnv1a.go
```diff
@@ -0,0 +1,39 @@
+package statsd
+
+const (
+	// FNV-1a
```
bai: Curious: is there any reason for not using https://golang.org/src/hash/fnv/fnv.go?s=1294:1319#L40?

arbll: Unfortunately, the way the standard hash API is exposed automatically leads to dynamic allocations. This is not desirable in our case since it's pretty common to have calls to these library methods in application hot paths.

kevinconaway:
Thoughts on allowing clients to control the "sharding"?  Perhaps something like

```go
client := statsd.Client

for range workerCount {
   go func() {
      stats := client.NewClient()
      //....
   }()
}
```

Each "sub" client would then be independent from the others and could buffer / append metrics internally before sending them to to be delivered.

I'm concerned that hashing every metric key is going to add unnecessary overhead for the scenarios where we submit metrics in tight loops

arbll:
@kevinconaway I could add a fast path for the case where we only configured one shard. That should work for such use cases right ?

statsd/worker.go
```diff
@@ -0,0 +1,80 @@
...
+func newWorker(pool *bufferPool, sender *sender) *worker {
+	return &worker{
+		pool:   pool,
+		sender: sender,
+		buffer: pool.borrowBuffer(),
+	}
+}
+
+func (w *worker) processMetric(m metric) error {
+	if w.shouldSample(m.rate) {
```
truthbk: I believe this should be `if !w.shouldSample(m.rate) {}`?

statsd/client_test.go
```diff
@@ -150,16 +150,16 @@ func TestBufferedClient(t *testing.T) {
...
-	client.Decr("dc", nil, 1)
-	client.Count("cc", 1, nil, 1)
-	client.Gauge("gg", 10, nil, 1)
-	client.Histogram("hh", 1, nil, 1)
-	client.Distribution("dd", 1, nil, 1)
-	client.Timing("tt", dur, nil, 1)
-	client.Set("ss", "ss", nil, 1)
-
-	client.Set("ss", "xx", nil, 1)
+	client.Incr("ab", nil, 1)
```
truthbk: although you do it on the benchmark test, would we want to have a different test with more random metric naming to ensure we test the actual sharding of the metrics to the different workers?

statsd/worker.go
hush-hush: The logic is right but the naming is confusing, editing.

statsd/statsd.go
```diff
@@ -395,44 +382,12 @@ func (c *Client) namespace() string {
...
-	c.Lock()
-	var err error
-	if err = c.writeMetricUnsafe(m); err == errBufferFull {
-		c.flushUnsafe()
-		err = c.writeMetricUnsafe(m)
-	}
-	c.Unlock()
-	return err
+	h := hashString32(m.name)
+	return c.bufferShards[h%uint32(len(c.bufferShards))].processMetric(m)
```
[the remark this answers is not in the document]
hush-hush: The `fastrange` is fair only if we have a wide spread of value. We might have similar value in our case since many metrics will share the majority of there name.

For example, running some very simple test with our benchmark (we have GOMAXPROC different metric name, so in my case `test.metric1` to `test.metric8`):

This produce this spread with `%`:
```
modulo test.metric1 -> shard 28
modulo test.metric2 -> shard 21
modulo test.metric3 -> shard 2
modulo test.metric4 -> shard 3
modulo test.metric5 -> shard 16
modulo test.metric6 -> shard 9
modulo test.metric7 -> shard 22
modulo test.metric8 -> shard 7
```

This produce this spread with `fastrange`:
```
fastrange test.metric1 -> shard 14
fastrange test.metric2 -> shard 14
fastrange test.metric3 -> shard 14
fastrange test.metric4 -> shard 14
fastrange test.metric5 -> shard 14
fastrange test.metric6 -> shard 14
fastrange test.metric7 -> shard 14
fastrange test.metric8 -> shard 15
```

Having fully randomized metrics name give the same spread than the modulo but does not speed things up much:

fully randomize metric name + modulo:
```
BenchmarkStatsdUDP-8   	22229242	        52.7 ns/op
BenchmarkStatsdUDS-8   	21336699	        55.6 ns/op
```

fully randomize metric name + fast range:
```
BenchmarkStatsdUDP-8   	23752176	        51.0 ns/op
BenchmarkStatsdUDS-8   	24448677	        50.9 ns/op
```

statsd/client_test.go
hush-hush: Not sure it's worth it:
- Testing with arbitrary value is not very valuable I think
- Testing with random value will cause the test to be flaky

Since it's not part of the public API I don't think unit test should cover this part. The  benchmark are better suited for this I think.

truthbk: You're absolutely right, randomizing the metric name would make the test unreliable. Sounds good to just do this on the benchmark test.

statsd/statsd.go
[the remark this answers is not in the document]
truthbk: modulo it is! But thanks for doing this. It's interesting how we could not reproduce at all the kind of gains `fastrange` claims. Just curious how many buckets did we have here?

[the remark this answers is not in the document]
hush-hush: The default it 32 (https://github.com/DataDog/datadog-go/pull/108/files#diff-4471ec43bb44996af266fa51c1a2acf3R22)

PR #120, opened 2020-01-13
fix a race between flushing and writing

arbll:
### What does this PR do

When calling `sender.flush()` one datagram could be in the process of being written by `sender.sendLoop()`. When flushing just before exiting the app one datagram could be lost because it was still in the process of being handled by `sendLoop`.

I am not 100% sure we need this as a call to `client.Close()` will flush and make sure `sendLoop` completes. This issue can be viewed as a missing `Close` call by the user.


arbll:
Looks like this was implemented in https://github.com/DataDog/datadog-go/pull/163 instead

Issue #123, opened 2020-02-06
Deadlock in statsd.Client.Close()


danp60:
I've encountered what I believe to be a deadlock in the `Close()` logic. When the main goroutine is closing, the mutex for the `Client` is acquired, we send a signal to all other goroutines to stop, then we wait for those goroutines to stop: https://github.com/DataDog/datadog-go/blob/master/statsd/statsd.go#L547-L555
``` golang
func (c *Client) Close() error {
	if c == nil {
		return ErrNoClient
	}
	c.Lock() // <- acquire the lock here
	defer c.Unlock()
	select {
	case <-c.stop:
		return nil
	default:
	}
	close(c.stop)
	c.wg.Wait()
```

However, in the `watch()` goroutine, its possible for the ticker to fire _after_ the main thread has acquired the mutex, but before the `stop` channel is closed. This means that the `watch()` goroutine will also try and acquire the same mutex here: https://github.com/DataDog/datadog-go/blob/master/statsd/statsd.go#L334 and block.
``` golang
func (c *Client) watch() {
	ticker := time.NewTicker(c.flushTime)

	for {
		select {
		case <-ticker.C:
			c.Lock() // <- This blocks because the main goroutine holds the lock
```

Since the main goroutine waits for the `watch()` goroutine to exit here: https://github.com/DataDog/datadog-go/blob/master/statsd/statsd.go#L555, we have a deadlock.

``` golang
func (c *Client) Close() error {
	if c == nil {
		return ErrNoClient
	}
	c.Lock() // <- acquire the lock here
	defer c.Unlock()
	select {
	case <-c.stop:
		return nil
	default:
	}
	close(c.stop)
	c.wg.Wait() # <- This deadlocks because `watch()` is waiting for c.Lock() to be released
```

hush-hush:
This has been fixed in `3.4.1`

jared2501:
Thank you!

PR #124, opened 2020-02-06
Fix potential deadlock on Close()

danp60:
Fixes deadlock issue described in #123

### DIFF
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -199,6 +199,10 @@ type Client struct {
 	stop          chan struct{}
 	wg            sync.WaitGroup
 	sync.Mutex
+	closerLock    sync.Mutex
+
+	// used for deadlock test
+	debugDeadlock chan struct{}
 }
 
 // ClientMetrics contains metrics about the client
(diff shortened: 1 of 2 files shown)


jared2501:
ping @arbll @jmoiron - would love to get this change looked at. I ran into this issue separately.

statsd/statsd.go
```diff
@@ -331,6 +335,11 @@ func (c *Client) watch() {
 	for {
 		select {
 		case <-ticker.C:
+			// c.debugDeadlock is always nil except for the TestCloseDeadlock test.
+			if c.debugDeadlock != nil {
+				<-c.debugDeadlock
+			}
+
```
hush-hush: Could we instead overwrite the `NewTicker` function ? Like we do [here](https://github.com/DataDog/datadog-agent/blob/master/pkg/metadata/scheduler.go#L25).

That way the code logic is the same when running tests or not and we can control when the ticker triggers.

jared2501: @hush-hush - do you think you can just take this PR over in a separate branch and make the changes yourself? I am a datadog customer and rely on datadog-go to integrate my company's product into the datadog stack. This bug is preventing me from upgrading our dependency and taking advantage of new features.

hush-hush: After looking more into this I think it's fine as it is :+1:

danp60:
Apologies, I haven't had time to look at the test logic feedback. What do I need to do to get this merged upstream?

hush-hush:
Hi @danp60,

I cherry-picked your fix in #135. Thanks again for this fix ! This will be included in the next release.

PR #127, opened 2020-03-01
support newer go versions and fix how to run test command

grosser:
old command did not work and dependencies were not declared
```
go test
can't load package: package .: no Go files in ~/Code/tools/datadog-go
```


grosser:
it still works with modules disabled afaik (older ci builds did not fail)
I was unable to run tests locally (1.13) without that file though

suhasgaddam:
@remeh @hush-hush 
Any chance this can get merged because it does not break any backwards compatibility?

hush-hush:
Hi,

We've added go modules support in the new major.

PR #134, opened 2020-03-09
Adding a drop mode to the client

hush-hush:
# What does this PR do

This PR add a "drop mode" to the DogStatsD client.

The current implementation uses mutexes sharded by metric name. This is the
fastest way to accept metrics but can create lock contention when sending a
very high number of metrics per second from multiple goroutine.
The mutex sharding solve the use case of multiple goroutine sending different
metrics or just sending a normal number of points per second (in the hundreds of
thousand per second).

But, at a very high rate (millions of points per second or above), the
application will be slowed down by dogstatsd due to lock contention. The new
"drop mode" based on channel is slower than mutexes but will actually drop
metrics when the channel is fulled. This is a trade off between never slowing
down the application but drop point.

The new "drop mode" is disable by default and answer a very specific use case:
dropping point instead of being slowed down when sending the same metrics at a
very high rate from multiple goroutine. It should not be enabled by default.

### Different metrics

When sending different metrics from each goroutine the mutex are faster and
don't drop 80% of metrics. This highlight the fact that the drop mode should
only be used in very specific cases.

```
BenchmarkStatsdUDPDifferentMetricBlocking-8   	14167192	        73.7 ns/op	         0 %_dropRate
BenchmarkStatsdUDPDifferentMetricDropping-8   	16056422	       115 ns/op	        80.2 %_dropRate

BenchmarkStatsdUDSDifferentMetricBlocking-8   	17318366	        75.8 ns/op	         0 %_dropRate
BenchmarkStatsdUDSDifferentMetricDropping-8   	14119225	        93.5 ns/op	        84.8 %_dropRate
```

### Same metrics

When sending the same metrics from each goroutine, "drop mode" is 50% faster at
the cost of dropping a lot of points.

```
BenchmarkStatsdUDPSameMetricBLocking-8        	 5100489	       228 ns/op	         0 %_dropRate
BenchmarkStatsdUDPSameMetricDropping-8        	11213953	       104 ns/op	        92.8 %_dropRate

BenchmarkStatsdUDSSameMetricBLocking-8        	 4567035	       254 ns/op	         0 %_dropRate
BenchmarkStatsdUDSSameMetricDropping-8        	13040250	        88.3 ns/op	        93.1 %_dropRate
```

### internal Async Client

Our internal Async Client wrapper that implement "drop mode" above the Go
dogstatsd client is slower but had similar drop rates. We should be able to
remove that wrapper in favor of "drop mode".

```
BenchmarkStatsdUDPSameMetricDropping-8        	11213953	       104 ns/op	        92.8 %_dropRate
BenchmarkStatsdUDPSameMetricAsync-8           	 9906213	       120 ns/op	        89.0 %_dropRate

BenchmarkStatsdUDPDifferentMetricDropping-8   	16056422	       115 ns/op	        80.2 %_dropRate
BenchmarkStatsdUDPDifferentMetricAsync-8      	 8388811	       126 ns/op	        88.0 %_dropRate

BenchmarkStatsdUDSSameMetricDropping-8        	13040250	        88.3 ns/op	        93.1 %_dropRate
BenchmarkStatsdUDSSameMetricAsync-8           	10397424	       124 ns/op	        89.1 %_dropRate

BenchmarkStatsdUDSDifferentMetricDropping-8   	14119225	        93.5 ns/op	        84.8 %_dropRate
BenchmarkStatsdUDSDifferentMetricAsync-8      	11211751	       116 ns/op	        86.6 %_dropRate
```

### DIFF
--- a/statsd/options.go
+++ b/statsd/options.go
@@ -26,6 +26,10 @@ var (
 	DefaultWriteTimeoutUDS = 1 * time.Millisecond
 	// DefaultTelemetry is the default value for the Telemetry option
 	DefaultTelemetry = true
+	// DefaultReceivingingMode is the default behavior when sending metrics
+	DefaultReceivingMode = MutexMode
+	// DefaultChannelModeBufferSize is the default size of the channel holding incoming metrics
+	DefaultChannelModeBufferSize = 4096
 )
 
 // Options contains the configuration options for a client.
@@ -60,6 +64,28 @@ type Options struct {
 	// Telemetry is a set of metrics automatically injected by the client in the
 	// dogstatsd stream to be able to monitor the client itself.
 	Telemetry bool
+	// ReceiveMode determins the behavior of the client when receiving to many
+	// metrics. The client will either drop the metrics if its buffers are
+	// full (ChannelMode mode) or block the caller until the metric can be
+	// handled (MutexMode mode). By default the client will MutexMode. This
+	// option should be set to ChannelMode only when use under very high
+	// load.
+	//
+	// MutexMode uses a mutex internally which is much faster than
+	// channel but causes some lock contention when used with a high number
+	// of threads. Mutex are sharded based on the metrics name which
+	// limit mutex contention when goroutines send different metrics.
+	//
+	// ChannelMode: uses channel (of ChannelModeBufferSize size) to send
+	// metrics and drop metrics if the channel is full. Sending metrics in
+	// this mode is slower that MutexMode (because of the channel), but
+	// will not block the application. This mode is made for application
+	// using many goroutines, sending the same metrics at a very high
+	// volume. The goal is to not slow down the application at the cost of
+	// dropping metrics and having a lower max throughput.
+	ReceiveMode ReceivingMode
+	// ChannelModeBufferSize is the size of the channel holding incoming metrics
+	ChannelModeBufferSize int
 }
 
 func resolveOptions(options []Option) (*Options, error) {
@@ -74,6 +100,8 @@ func resolveOptions(options []Option) (*Options, error) {
 		SenderQueueSize:       DefaultSenderQueueSize,
 		WriteTimeoutUDS:       DefaultWriteTimeoutUDS,
 		Telemetry:             DefaultTelemetry,
+		ReceiveMode:           DefaultReceivingMode,
+		ChannelModeBufferSize: DefaultChannelModeBufferSize,
 	}
 
 	for _, option := range options {
@@ -165,6 +193,30 @@ func WithWriteTimeoutUDS(writeTimeoutUDS time.Duration) Option {
 func WithoutTelemetry() Option {
 	return func(o *Options) error {
 		o.Telemetry = false
+		return nil
+	}
+}
+
+// WithChannelMode will use channel to receive metrics
+func WithChannelMode() Option {
+	return func(o *Options) error {
+		o.ReceiveMode = ChannelMode
+		return nil
+	}
+}
+
+// WithMutexModeMode will use mutext to receive metrics
+func WithMutexMode() Option {
+	return func(o *Options) error {
+		o.ReceiveMode = MutexMode
+		return nil
+	}
+}
+
+// WithChannelModeBufferSize the channel buffer size when using "drop mode"
+func WithChannelModeBufferSize(bufferSize int) Option {
+	return func(o *Options) error {
+		o.ChannelModeBufferSize = bufferSize
 		return nil
 	}
 }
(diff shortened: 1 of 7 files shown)


statsd/statsd_benchmark_test.go
```diff
@@ -2,16 +2,22 @@ package statsd_test
...
+	"strconv"
+	"strings"
 	"sync/atomic"
 	"testing"
 
 	"github.com/DataDog/datadog-go/statsd"
 )
 
-func setupUDSClientServer(b *testing.B) (*statsd.Client, net.Listener) {
+var goMinorVersion, _ = strconv.Atoi(strings.Split(runtime.Version(), ".")[1])
```
olivielpeau: nit: I don't think this variable is actually used. I guess you wanted to use it to report the additional drop metric only on go1.13 and up?

To do so you could put this logic in a different test file behind the `go1.13` build tag, see https://golang.org/pkg/go/build/#hdr-Build_Constraints

statsd/statsd.go
```diff
@@ -101,6 +101,13 @@ const (
 	serviceCheck
 )
 
+type SendingMode int
+
+const (
+	BlockOnSend SendingMode = iota
```
arbll: Maybe too much of a nitpick but I find those names confusing:
- "Block" can be interpreted as blocking on network (neither mode will block on network)
- Both modes will drop metrics on send if queues get full
- Chan are not lock-free, the contention will just be lower

statsd/options.go
```diff
@@ -60,20 +64,43 @@ type Options struct {
...
+	// SendMode determins the behavior of the client when receiving to many
+	// metrics. The client will either drop the metrics if its buffers are
+	// full (DropOnSend mode) or block the caller until the metric can be
+	// handled (BlockOnSend mode). By default the client will block. This
+	// option should be set to DropOnSend only when use under very high
+	// load.
+	//
+	// BlockOnSend uses a mutex internally which is much faster than
+	// channel but causes some lock contention when used with a high number
+	// of goroutine. Mutex are sharded based on the metrics name which
```
arbll:
```suggestion
	// of threads. Mutex are sharded based on the metrics name which
```

statsd/options.go
```diff
@@ -26,6 +26,10 @@ var (
 	DefaultWriteTimeoutUDS = 1 * time.Millisecond
 	// DefaultTelemetry is the default value for the Telemetry option
 	DefaultTelemetry = true
+	// DefaultSendingMode is the default behavior when sending metrics
+	DefaultSendingMode = BlockOnSend
+	// DefaultDropSendingModeBufferSizer is the default size of the channel holding incoming metrics
+	DefaultDropSendingModeBufferSizer = 2048
```
arbll: Did you test the drop rate with a rate limited client ? We usually see about 16 metrics per UDP payload. Since the client currently holds 2048 buffers of ~16 metrics by default we will start dropping 16 times earlier

statsd/statsd_benchmark_test.go
hush-hush: You read my mistakes like an open book ;)

statsd/statsd.go
hush-hush: I struggled a bit on the naming. Do you have any proposition ? This is the best I could come up with.

statsd/options.go
hush-hush: Not sure I understand the comment. The `DefaultDropSendingModeBufferSizer` is for the input buffer before the worker (so before batching). That buffer holds up to 2048 single metrics at a time by default.

A rate limited client would drop less metrics since it would leave time to the workers to consume the metrics.

statsd/statsd.go
arbll: IMO we can use Mutex & (Non-Blocking) Channel in the naming. Like you said this is a very specialized thing that we don't expect regular people to use so I don't think we need to hide the technical details in the names

statsd/options.go
arbll: What I meant was that 2048 single metrics is much lower than what we have today in the next pipe. I am a bit worried it's going to fill up fast even at relatively low MPS. I mentioned a rate limiter because the current benchmarks only tells us the drop % when you fully saturate the client. I think it would be more interesting to get data at a few fixed rates.

hush-hush: I used this because that's what  we have internally in our wrapper around dogatstsd (that implement the channel mode).

We could bump it to `4096` but knowing that we have a channel per worker that means 32*2048 metrics. Of course if a a client sends the same metrics the worker sharding doesn't help. I'm worried of the default memory consumption of having a higher default. WDYT ?

hush-hush: Merging for now to unlock client side aggregation. We can change the default later.

PR #137, opened 2020-03-10
Add support for DD_ENV, DD_SERVICE, and DD_VERSION environment variables

jdgumz:
This PR adds support for:
- `DD_ENV`
- `DD_SERVICE`
- `DD_VERSION`

to be used by the statsd client at initialization. The first three are added as global tags for `{env, service, version}`, if they are set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,8 @@ env:
         fieldPath: metadata.uid
 ```
 
+* `DD_ENV`, `DD_SERVICE`, and `DD_VERSION` can be used by the statsd client to set `{env, service, version}` as global tags for all data emitted.
+
 ### Unix Domain Sockets Client
 
 Agent v6+ accepts packets through a Unix Socket datagram connection. Details about the advantages of using UDS over UDP are available in the [DogStatsD Unix Socket documentation](https://docs.datadoghq.com/developers/dogstatsd/unix_socket/). You can use this protocol by giving a `unix:///path/to/dsd.socket` address argument to the `New` constructor.
--- a/statsd/client_test.go
+++ b/statsd/client_test.go
@@ -390,9 +390,8 @@ func TestServiceChecks(t *testing.T) {
 	}
 }
 
-const entityIDEnvName = "DD_ENTITY_ID"
-
 func TestEntityID(t *testing.T) {
+	entityIDEnvName := "DD_ENTITY_ID"
 	initialValue, initiallySet := os.LookupEnv(entityIDEnvName)
 	if initiallySet {
 		defer os.Setenv(entityIDEnvName, initialValue)
--- a/statsd/client_test.go
+++ b/statsd/client_test.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"os"
 	"reflect"
+	"sort"
 	"strings"
 	"sync"
 	"testing"
@@ -14,12 +15,6 @@ import (
 
 	"github.com/DataDog/datadog-go/statsd"
 	"github.com/stretchr/testify/assert"
-)
-
-// Client-side entity ID injection for container tagging
-const (
-	entityIDEnvName = "DD_ENTITY_ID"
-	entityIDTagName = "dd.internal.entity_id"
 )
 
 var dogstatsdTests = []struct {
@@ -395,6 +390,8 @@ func TestServiceChecks(t *testing.T) {
 	}
 }
 
+const entityIDEnvName = "DD_ENTITY_ID"
+
 func TestEntityID(t *testing.T) {
 	initialValue, initiallySet := os.LookupEnv(entityIDEnvName)
 	if initiallySet {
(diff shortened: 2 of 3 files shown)


statsd/statsd.go
```diff
@@ -278,6 +286,23 @@ func newWithWriter(w statsdWriter, o *Options, writerName string) (*Client, erro
 		c.Tags = append(c.Tags, entityTag)
 	}
 
+	// Inject DD_ENV, DD_SERVICE, and DD_VERSION as global tags if present.
+	env := os.Getenv(ddEnvName)
+	if env != "" {
+		envTag := fmt.Sprintf("%s:%s", ddEnvTagName, env)
```
hush-hush: We could make this generic to ease adding support to more env tags.

something like:
```go
var envTagsMapping := map[string]string{
  "DD_ENV":     "env".
  "DD_SERVICE": "service",
  "DD_VERSION": "version",
}

// [...]

for envName, tagName := range envTagsMapping {
    if value := os.Getenv(envName); value != "" {
        c.Tags = append(c.Tags, fmt.Sprintf("%s:%s", tagName, value))
    }
}
```

WDYT ?

README.md
```diff
@@ -56,6 +56,8 @@ env:
         fieldPath: metadata.uid
 ```
 
+* `DD_ENV`, `DD_SERVICE`, and `DD_VERSION` can be used by the statsd client to globally set `{env, service, version}` as tags for all data emitted. The values of these tags will not override any other configured tags for `{env, service, version}`.
```
hush-hush: I think we should rename the `will not override` since the code will actually send multiple value for the same tag if both set in the ENV and in the `Tags`.

statsd/client_test.go
```diff
@@ -437,6 +445,118 @@ func TestEntityID(t *testing.T) {
...
+	os.Setenv(ddEnvName, "prod")
+	os.Setenv(ddServiceName, "dog")
+	os.Setenv(ddVersionName, "abc123")
+	client, err := statsd.New("localhost:8125")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(client.Tags) != 3 {
+		t.Errorf("Expecting three tags, got %d", len(client.Tags))
+	}
```
hush-hush: Could we use `assert` here ?

```go
assert.Len(t, client.Tags, 3)
```

Same for the rest.

statsd/client_test.go
```diff
@@ -437,6 +445,118 @@ func TestEntityID(t *testing.T) {
...
+	if client.Tags[0] != "env:prod" {
+		t.Errorf("Bad tag value, got %s", client.Tags[0])
+	}
+	if client.Tags[1] != "service:dog" {
+		t.Errorf("Bad tag value, got %s", client.Tags[1])
+	}
+	if client.Tags[2] != "version:abc123" {
+		t.Errorf("Bad tag value, got %s", client.Tags[2])
+	}
```
hush-hush: We could test once with a list.

```go
assert.Equal(t, client.Tags, []string{"env:prod", "service:dog", "version:abc123"})
```

statsd/statsd.go
jdgumz: This is a good idea, I will make the change!

README.md
jdgumz: :+1: I've updated the docs

statsd/client_test.go
jdgumz: Sure thing; I didn't see much use of `assert` in the other test cases but there's no reason not to use it here.

statsd/statsd.go
```diff
@@ -271,11 +286,25 @@ func newWithWriter(w statsdWriter, o *Options, writerName string) (*Client, erro
...
+		if tag == "" {
+			continue
+		}
+		parts := strings.Split(tag, ":")
+		if len(parts) == 1 {
+			c.Tags = append(c.Tags, parts[0])
+		}
+		if len(parts) == 2 {
+			c.Tags = append(c.Tags, fmt.Sprintf("%s:%s", parts[0], parts[1]))
+		}
```
hush-hush: Why split ? What is the value of a tag contains `:` ?

jdgumz: Good call, we do support `:` in tag values.

I've updated the logic and test cases accordingly.

statsd/statsd.go
```diff
@@ -83,12 +83,28 @@ traffic instead of UDP.
 */
 const UnixAddressPrefix = "unix://"
 
-// Client-side entity ID injection for container tagging
+/*
+Client-side entity ID injection for container tagging.
+*/
```
hush-hush: No need to change the comment

statsd/statsd.go
```diff
@@ -83,12 +83,28 @@ traffic instead of UDP.
...
 const (
 	entityIDEnvName = "DD_ENTITY_ID"
 	entityIDTagName = "dd.internal.entity_id"
 )
```
hush-hush: I don't think we need to keep those variables around, we could hardcode them in the `ddEnvTagsMapping` like the other ones.

PR #139, opened 2020-03-18
[POC] Client side aggregation.

hush-hush:
First, early try of client side aggregation.

For now this looks promising. We're testing both best case scenarion (1 metric sampled a lot) and worst case scenario (a lot of metrics sampled only once).

```
goos: linux
goarch: amd64
pkg: github.com/DataDog/datadog-go/statsd
BenchmarkStatsdUDPSameMetricBLocking-8                   	 5984832	       218 ns/op	         0 %_dropRate
BenchmarkStatsdUDPSameMetricDropping-8                   	13658887	        85.3 ns/op	        93.3 %_dropRate
BenchmarkStatsdUDPSameMetricBLockingAggregation-8        	13163277	        81.6 ns/op	         0 %_dropRate
BenchmarkStatsdUDPSameMetricDroppingAggregation-8        	15646678	        74.6 ns/op	         0 %_dropRate
BenchmarkStatsdUDPDifferentMetricBlocking-8              	17076426	        64.3 ns/op	         0 %_dropRate
BenchmarkStatsdUDPDifferentMetricDropping-8              	17353012	        66.9 ns/op	        83.4 %_dropRate
BenchmarkStatsdUDPDifferentMetricBlockingAggregation-8   	14357695	        81.4 ns/op	         0 %_dropRate
BenchmarkStatsdUDPDifferentMetricDroppingAggregation-8   	17213880	        68.9 ns/op	         0 %_dropRate
BenchmarkStatsdUDSSameMetricBLocking-8                   	 6019815	       214 ns/op	         0 %_dropRate
BenchmarkStatsdUDSSameMetricDropping-8                   	16993280	        81.6 ns/op	        93.4 %_dropRate
BenchmarkStatsdUDSSameMetricBLockingAggregation-8        	16397464	        85.0 ns/op	         0 %_dropRate
BenchmarkStatsdUDSSameMetricDroppingAggregation-8        	15898207	        68.9 ns/op	         0 %_dropRate
BenchmarkStatsdUDPSifferentMetricBlocking-8              	15929932	        64.1 ns/op	         0 %_dropRate
BenchmarkStatsdUDSDifferentMetricDropping-8              	15504646	        64.6 ns/op	        82.0 %_dropRate
BenchmarkStatsdUDPSifferentMetricBlockingAggregation-8   	14955110	        76.3 ns/op	         0 %_dropRate
BenchmarkStatsdUDSDifferentMetricDroppingAggregation-8   	16314648	        69.6 ns/op	         0 %_dropRate
```


statsd/statsd.go
```diff
@@ -288,6 +289,10 @@ func newWithWriter(w statsdWriter, o *Options, writerName string) (*Client, erro
 		Tags:          o.Tags,
 		telemetryTags: []string{clientTelemetryTag, clientVersionTelemetryTag, "client_transport:" + writerName},
 	}
+	if o.Aggregation {
+		c.agg = newAggregator(&c)
+		c.agg.start(o.AggregationFlushInterval)
```
remeh: I think that's better to use the `go` keyword here instead of into `aggregator.start`, making it obvious _here_ that it'll be running concurrently.

hush-hush: I don't agree actually. I think the aggregator should not expose it's internal behavior to the outside. The caller should not have to know how to start the aggregator (with a goroutine or not). From a user point of view, you create the `aggregator` and then it's `start` and `stop`.

The complexity should be encapsulate inside the class. WDYT ?

remeh: Got your point!

> The caller should not have to know how to start the aggregator (with a goroutine or not).

That's where I differ because I think it is the user of the aggregator which is deciding to run it concurrently. (I'm not trying to convince you, just sharing my thoughts, I'm 100% fine with this solution 👍)

hush-hush: When would you not want to run it concurrently ? I don't see the use case. By design the aggregator needs to run concurrently (so we can sample metrics on one side and aggregate on the other side).

statsd/aggregator.go
```diff
@@ -0,0 +1,149 @@
...
+	gauge := newGaugeMetric(name, value, tags, rate)
+
+	a.gaugesM.Lock()
+	a.gauges[context] = gauge
+	a.gaugesM.Unlock()
+	return nil
+}
+
+func (a *aggregator) set(name string, value string, tags []string, rate float64) error {
+	context := name + strings.Join(tags, "")
```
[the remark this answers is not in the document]
hush-hush: good point !

statsd/aggregator.go
```diff
@@ -0,0 +1,149 @@
...
+	for _, c := range counts {
+		metrics = append(metrics, c.flushUnsafe())
+	}
+
+	return metrics
+}
+
+func (a *aggregator) count(name string, value int64, tags []string, rate float64) error {
+	context := name + strings.Join(tags, "")
+	a.countsM.RLock()
```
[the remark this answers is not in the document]
hush-hush: Indeed, we should not release the read lock until we sampled the metric. This should be fine as we won't slow down the sampling.

statsd/metrics.go
```diff
@@ -0,0 +1,124 @@
...
+		name: name,
+		tags: tags,
+		rate: rate,
+	}
+	set.data[value] = struct{}{}
+	return set
+}
+
+func (s *setMetric) sample(v string) {
+	s.Lock()
```
ogaca-dd, line 100: As `sample` is used inside a block  `a.countsM.RLock()`  / `a.countsM.RUnlock()`, `s.Lock()` is not required anymore.

statsd/statsd.go
remeh: Yup that always needs concurrency; it is just that you don't know from here that starting the aggregator is non-blocking. Definitely not an "issue" just a standpoint, resolving this thread!

statsd/metrics.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// Sets are aggregated on the agent side too. We flush the keys so a set from
+// multiple application can be correctly aggregated on the agnet side.
+func (s *setMetric) flushUnsafe() []metric {
+	if len(s.data) == 0 {
+		return nil
+	}
+
+	metrics := []metric{}
```
remeh, line 112: As you know the size, you may use the assignation form to avoid re-allocation / resize of the slice on the flight.

```
metrics := make([]metric, len(s.data))
for (i := 0; i < len(s.data); i++) {
  metrics[i] = metric{ metricType: set, ... }
}
```

statsd/statsd.go
```diff
@@ -474,22 +461,44 @@ func (c *Client) send(m metric) error {
...
+		return ErrNoClient
+	}
+	atomic.AddUint64(&c.metrics.TotalMetrics, 1)
+	return c.send(metric{metricType: histogram, name: name, fvalue: value, tags: tags, rate: rate})
 }
 
 // Distribution tracks the statistical distribution of a set of values across your infrastructure.
 func (c *Client) Distribution(name string, value float64, tags []string, rate float64) error {
-	return c.addMetric(metric{namespace: c.namespace(), globalTags: c.globalTags(), metricType: distribution, name: name, fvalue: value, tags: tags, rate: rate})
+	if c == nil {
```
remeh, line 497: 👌

statsd/metrics.go
hush-hush: I think it is to avoid concurrent write to the map, no ?

PR #141, opened 2020-03-30
Add go modules support

suhasgaddam:


suhasgaddam:
Closed in favor of #127 which does the same thing

PR #143, opened 2020-04-10
report telemetry tags with the same set as global tags

chrisleavoy:
it seems like an obvious oversight that this wasn't done already... its basically impossible to track these metrics against where they come from without applying some sort of tags to identify the origin.

### DIFF
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -301,9 +301,8 @@ func newWithWriter(w statsdWriter, o *Options, writerName string) (*Client, erro
 	w.SetWriteTimeout(o.WriteTimeoutUDS)
 
 	c := Client{
-		Namespace:     o.Namespace,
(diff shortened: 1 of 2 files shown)


statsd/statsd.go
```diff
@@ -288,6 +287,8 @@ func newWithWriter(w statsdWriter, o *Options, writerName string) (*Client, erro
 		}
 	}
 
+	c.telemetryTags = append(o.Tags, clientTelemetryTag, clientVersionTelemetryTag, "client_transport:" + writerName)
```
hush-hush, line 290: We should use `c.Tags` instead of `o.Tags` to also use environment tags.

hush-hush:
I update the PR and merged it.

Thanks @chrisleavoy for this !

chrisleavoy:
👍 thanks @hush-hush

PR #147, opened 2020-04-28
Adding a clone method to the client

hush-hush:
This allows libraries to add options to the main dogstatsd client
without impacting the main application.

### DIFF
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -217,6 +217,8 @@ type Client struct {
 	closerLock    sync.Mutex
 	receiveMode   ReceivingMode
 	agg           *aggregator
+	options       []Option
+	addrOption    string
 }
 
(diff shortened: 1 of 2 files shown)


statsd/statsd.go
```diff
@@ -280,6 +287,15 @@ func NewWithWriter(w statsdWriter, options ...Option) (*Client, error) {
 	return newWithWriter(w, o, "custom")
```
ogaca-dd, line 287: Maybe fields `options` and `addrOption` can also be set in `NewWithWriter`?

hush-hush: The issue with with `NewWithWriter` is that both client would share the same writer. This means that Closing one client would close the transport of the other one too.
That's why we raise an error when cloning a client created with `NewWithWriter`.

PR #150, opened 2020-06-06
Add a top-level go.mod for go module users

### LINKED ISSUE #119: Go Modules Support
Currently there are no `go.mod` and `go.sum` files in the repo. `go mod` is currently adding `v<version>-incompatible` tags to `go.mod` files but they are prone to breaking whenever we `go get -u` the dependency in modules mode.

It'd be very helpful to have Go Modules support to make dependency management more deterministic.

evantorrie:
Fixes #119

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -20,4 +20,3 @@ Example Usage:
 statsd is based on go-statsd-client.
 */
 package datadog_go
-
--- a/doc.go
+++ b/doc.go
@@ -1,0 +1,23 @@
+/*
+Package statsd provides a Go dogstatsd client. Dogstatsd extends the popular statsd,
+adding tags and histograms and pushing upstream to Datadog.
+
+Refer to http://docs.datadoghq.com/guides/dogstatsd/ for information about DogStatsD.
+
+Example Usage:
+
+    // Create the client
+    c, err := statsd.New("127.0.0.1:8125")
+    if err != nil {
+        log.Fatal(err)
+    }
+    // Prefix every metric with the app name
+    c.Namespace = "flubber."
+    // Send the EC2 availability zone as a tag with every metric
+    c.Tags = append(c.Tags, "us-east-1a")
+    err = c.Gauge("request.duration", 1.2, nil, 1)
+
+statsd is based on go-statsd-client.
+*/
+package datadog_go
+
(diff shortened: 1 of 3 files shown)


doc.go
```diff
@@ -0,0 +1,22 @@
...
+    // Create the client
+    c, err := statsd.New("127.0.0.1:8125")
+    if err != nil {
+        log.Fatal(err)
+    }
+    // Prefix every metric with the app name
+    c.Namespace = "flubber."
+    // Send the EC2 availability zone as a tag with every metric
+    c.Tags = append(c.Tags, "us-east-1a")
+    err = c.Gauge("request.duration", 1.2, nil, 1)
```
hush-hush, line 18: Could we use the new API (this one will be deprecated in the next major):

```go
    // Create the client
￼    c, err := statsd.New("127.0.0.1:8125".
                                 ￼    // Prefix every metric with the app name
                                     statsd.WithNamespace("fluber."),
                                 ￼    // Send the EC2 availability zone as a tag with every metric
                                     statsd.WithTags([]string{"us-east-1a"}))
￼    if err != nil {
￼        log.Fatal(err)
￼    }
￼    err = c.Gauge("request.duration", 1.2, nil, 1)
    c.Close()
```

doc.go
```diff
@@ -0,0 +1,22 @@
+/*
+Package statsd provides a Go dogstatsd client. Dogstatsd extends the popular statsd,
+adding tags and histograms and pushing upstream to Datadog.
```
hush-hush, line 3:
```suggestion
Package statsd provides a DogStatsD client in Go. DogStatsD extends the popular statsd protocol by
adding Datadog specific type (events, service_check. distributions ...).
```

go.sum
```diff
@@ -0,0 +1,14 @@
+github.com/DataDog/datadog-go v3.7.1+incompatible h1:HmA9qHVrHIAqpSvoCYJ+c6qst0lgqEhNW6/KwfkHbS8=
+github.com/DataDog/datadog-go v3.7.1+incompatible/go.mod h1:LButxg5PwREeZtORoXG3tL4fMGNddJ+vMq1mwgfaqoQ=
```
hush-hush, line 2: I don't think we need those

doc.go
```diff
@@ -0,0 +1,22 @@
+/*
```
hush-hush, line 1: This kind of duplicate the readme and example. I'm not sure we need this.

hush-hush (changes requested):

hush-hush:
HI @evantorrie ,

We've added go modules support in the new major (`v5`).

I'm closing this PR. Feel free to open a new one if needed.

PR #151, opened 2020-06-06
Run with GOARCH=386 to test 32-bit code generation

evantorrie:
* Add a step where we compile with `GOARCH=386` to test 32-bit code generation. Should consider also adding a compile on non-Intel (i.e. ARM) architecture.

   * Eliminate very old versions of go - Go team officially supports only the last 2 major versions (i.e. back to go-1.12 at this point)


evantorrie:
This needs to wait for #152 to be merged before it will pass on a 386 build.

hush-hush:
Hi @evantorrie,

Thanks for opening this PR too but we went with #156 to fix this issue (which add both arm and 32bits). The patch should be released today.

PR #152, opened 2020-06-06
Ensure 8 byte alignment of 64-bit atomics

evantorrie:
This adds checks to ensure that any fields that are used as 64-bit atomics, i.e. `atomic.AddUint64(..)` are aligned on 8 byte boundaries as required for the 386 architecture.

See https://go101.org/article/memory-layout.html and open-telemetry/opentelemetry-go#341

Likely this will also address #142


hush-hush:
Hi @evantorrie,

Thanks for opening this PR but we went with #156 to fix this issue. The patch should be released today.

PR #61, opened 2018-08-07
[UDP] add some logic to recover from DNS changes

truthbk:
This is still in the works, just a proposed approach to automatically address changes to underlying IP addresses when there's DNS resolution - could come in handy in dynamic network environments.

PR #153, opened 2020-06-12
UDP reconnect

radykal-com:
Hi, I need this feature, so started with the code from #61 , but replaced the update channel with a RWMutex, because there was an edge case with the channel that would cause to always send to an old IP address if the IP address changed more frequently than metrics Writes due to the goroutine detecting the DNS change being blocked until the next write occurs.

### DIFF
--- a/statsd/udp.go
+++ b/statsd/udp.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"sync"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


colincromar:
@truthbk Any intention on merging this PR? This is functionality that would greatly help myself and my team.

radykal-com:
@colincromar we are using a custom binary compiled from this branch: https://github.com/stayforlong/datadog-go/tree/udp-reconnect 

It has been running for months just fine. Datadog seems not to be very active in this repo...

PR #157, opened 2020-06-25
Adding the option to send telemetry to another endpoint

hush-hush:
New option `WithTelemetryAddr` support UDP  and UDS like the rest of the client.

### DIFF
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -62,21 +62,6 @@ agent configuration file datadog.yaml.
 const DefaultMaxAgentPayloadSize = 8192
 
 /*
-TelemetryInterval is the interval at which telemetry will be sent by the client.
-*/
-const TelemetryInterval = 10 * time.Second
-
-/*
-clientTelemetryTag is a tag identifying this specific client.
-*/
-var clientTelemetryTag = "client:go"
-
-/*
-clientVersionTelemetryTag is a tag identifying this specific client version.
-*/
-var clientVersionTelemetryTag = "client_version:3.7.2"
-
-/*
 UnixAddressPrefix holds the prefix to use to enable Unix Domain Socket
 traffic instead of UDP.
 */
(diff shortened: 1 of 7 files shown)


statsd/statsd.go
```diff
@@ -321,8 +291,8 @@ func newWithWriter(w statsdWriter, o *Options, writerName string) (*Client, erro
 		}
 	}
 
-	c.telemetryTags = append(c.Tags, clientTelemetryTag, clientVersionTelemetryTag, "client_transport:"+writerName)
-
+	// FIXME: The agent has a performance pitfall preventing us from using better defaults here.
+	// Once it's fixed, use `DefaultMaxAgentPayloadSize` and `DefaultUDSBufferPoolSize` instead.
```
remeh, line 295: Because the code branches has changed, we don't really get anymore that we're speaking about the UDS defaults.
We should probably revamp the comment mentioning it, e.g.:

```
// FIXME: The agent has a performance pitfall preventing us from using better defaults for UDS,
// this is why we fallback on UDP defaults even in UDS mode.
// Once it's fixed, use `DefaultMaxAgentPayloadSize` and `DefaultUDSBufferPoolSize` instead for UDS.
```

statsd/telemetry.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Could not resolve telemetry address: %v", err)
+	}
+
+	t := NewTelemetryClient(c, transport)
+
+	// creating a custom sender/worker with 1 worker in
+	// mutex mode for the telemetry that share the same
+	// bufferPool
+	t.sender = newSender(telemetryWriter, DefaultUDPBufferPoolSize, pool)
```
remeh, line 51: I would probably add a FIXME about this default value as well. Or even the code

if (udp) {
  poolSize = default UDP
} else {
  poolSize = default UDP // FIXME due to performance pitfall, we're using UDP defaults for UDS.
}

statsd/statsd.go
```diff
@@ -333,31 +303,39 @@ func newWithWriter(w statsdWriter, o *Options, writerName string) (*Client, erro
...
+	if o.TelemetryAddr == "" {
+		c.telemetry = NewTelemetryClient(&c, writerName)
+	} else {
+		var err error
+		c.telemetry, err = NewTelemetryClientWithCustomAddr(&c, writerName, o.TelemetryAddr, bufferPool)
+		if err != nil {
+			return nil, err
 		}
-	}()
+	}
```
remeh, line 334: Don't we want do that in the next `if` line 336 `if o.Telemetry {`?

PR #158, opened 2020-07-01
Adding telemetry on the number of aggregated contexts

hush-hush:
Adding telemetry on the number of aggregated contexts

### DIFF
--- a/statsd/aggregator.go
+++ b/statsd/aggregator.go
@@ -13,19 +13,24 @@ type (
 )
 
 type aggregator struct {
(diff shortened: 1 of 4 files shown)


hush-hush:
I actually would be 2 contexts in the aggregator (one in the gauge map and one in the count map) and would end up in 2 metrics being sent over the wire. So I think counting 2 context is right.

I agree we don't want to support this use-case as it's a "wrong" usage of dogstatsd (the last "gauge" would nullify all the counter call).

PR #161, opened 2020-07-23
address race conditions

cyx:
## Description

This PR addresses the race conditions which are reproduced directly in
the tests introduced.

These race conditions were discovered by doing integration tests with a
service we're working with and running `-race` on that, while at the
same time having the datadog client running in the background.


cyx:
Just tested the latest release and there are indeed no more race problems. I'll just close this since I don't really have strong opinions about these.

Thanks!

PR #163, opened 2020-08-11
Fix incomplete flush in the sender when flushing the client

hush-hush:
In some case, the sender 'sendLoop' would pop a buffer from its queue
as we are calling its 'flush' method. Since the input queue is empty it
would return instantly without waiting for the current buffer own by the
sender to be sent. This would result in partial flush.

We now synchronize 'flush' and 'sendLoop' to make sure flushing the
client will send everything to the wire. The side effect of this is that
the workers are lock for longer as we keep them locked when flushing the
sender.

### DIFF
--- a/statsd/sender.go
+++ b/statsd/sender.go
@@ -28,20 +28,22 @@ type SenderMetrics struct {
 }
 
 type sender struct {
-	transport statsdWriter
-	pool      *bufferPool
-	queue     chan *statsdBuffer
-	metrics   *SenderMetrics
-	stop      chan struct{}
+	transport   statsdWriter
+	pool        *bufferPool
+	queue       chan *statsdBuffer
+	metrics     *SenderMetrics
+	stop        chan struct{}
+	flushSignal chan struct{}
 }
 
 func newSender(transport statsdWriter, queueSize int, pool *bufferPool) *sender {
 	sender := &sender{
-		transport: transport,
-		pool:      pool,
-		queue:     make(chan *statsdBuffer, queueSize),
-		metrics:   &SenderMetrics{},
-		stop:      make(chan struct{}),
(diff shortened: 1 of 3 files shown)


hush-hush:
Indeed the `ChannelMode` will not be fully flushed. The goal of it being to decouple sampling metrics and receiving it in the workers so we never block the app. Right now, we have no way to fully flush the input channel without blocking the calling app, which would defeat the main purpose of `ChannelMode`. I think the current behavior is expected but we might want to improve the `Flush` documentation to make it clear.

For the `flush message` idea, I don't think it would really work with multiple workers as we don't control which one will pull the message from the queue. I'm not sure the overhead of synchronizing the workers is worth it just for `ChannelMode`.

statsd/sender.go
```diff
@@ -90,16 +96,30 @@ func (s *sender) flushTelemetryMetrics() SenderMetrics {
...
 	}
 }
 
 func (s *sender) flush() {
+	// At that point we know that the workers are paused (the statsd client
+	// will pause them before calling sender.flush()).
+	// But we have to wait for sendLoop to be done with
+	// it's current buffer, it any.
+	s.flushLock.Lock()
+	defer s.flushLock.Unlock()
```
olivielpeau, line 121: how about using a semaphore instead of using a lock, i.e.: `flush` sends a message on a dedicated channel on which `sendLoop` listens in its main loop. When `sendLoop` receives the message, it reads everything that's left in the buffer (guaranteed to complete since the workers are paused), signals back to `flush` that it's done, and goes back to its main loop. `flush` then returns.

I think that would allow to get rid of the `time.After` logic.

truthbk: +1

PR #165, opened 2020-08-14
Drop support for EOL Go version 1.8 to 1.11

hush-hush:
Drop support for EOL Go version 1.8 to 1.11.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,25 +5,32 @@ arch:
   - arm64
 
 go:
-  - 1.8
-  - 1.9
-  - 1.10.x
-  - 1.11.x
-  - 1.12.x
   - 1.13.x
   - 1.14.x
+  - 1.15.x
 
-# test 32bit build too.
 jobs:
   include:
+    # test 32bit build too.
+    - arch: amd64
+      go: 1.15.x
+      script:
+        - GOARCH=386 go test -v ./...
     - arch: amd64
       go: 1.14.x
       script:
         - GOARCH=386 go test -v ./...
     - arch: amd64
-      go: 1.8
+      go: 1.13.x
       script:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -5,25 +5,32 @@ arch:
...
       script:
         - GOARCH=386 go test -v ./...
     - arch: amd64
-      go: 1.8
+      go: 1.13.x
       script:
         - GOARCH=386 go test -v ./...
+    # testing packages are no longer compatible with 1.12. We only test that we
+    # can build.
+    - go: 1.12.x
```
ogaca-dd, line 29: As far as I understand, go version from 1.8 to 1.11 are dropped because of an issue in unit tests. Is there a reason to keep 1.12 in this case?

hush-hush: Yes, 1.12.x is still supported by the APM product. Since it's only our tests that can't run with 1.12 we still check that we compile.

.travis.yml
```diff
@@ -5,25 +5,32 @@ arch:
...
 
 go:
-  - 1.8
-  - 1.9
-  - 1.10.x
-  - 1.11.x
-  - 1.12.x
   - 1.13.x
   - 1.14.x
+  - 1.15.x
```
ogaca-dd, line 11: As a test is run for go 1.12.x, it is OK to remove `-1.12.x` from here?

ogaca-dd (approved):
LGTM, just having a question.

.travis.yml
hush-hush: Yes, it's add a new test case. You can see it ran here: https://github.com/DataDog/datadog-go/pull/165/checks?check_run_id=986515245

PR #169, opened 2020-10-09
Adding a DevMode to the client

hush-hush:
## What this PR do

Adding a `DevMode` to the client.

The new mode will send more telemetry metrics allowing users to more easily troubleshoot issues and usage of the client.

Also,  this makes the telemetry creator functions private. Those function should not have been public in the first place and there signature should not be part of the public API we support. Users shouldn't have any reason to use them as they can control everything through the client options.

### DIFF
--- a/statsd/aggregator.go
+++ b/statsd/aggregator.go
@@ -14,7 +14,10 @@ type (
 )
 
 type aggregator struct {
-	nbContext int32
+	nbContext      int32
+	nbContextGauge int32
+	nbContextCount int32
+	nbContextSet   int32
 
 	countsM sync.RWMutex
 	gaugesM sync.RWMutex
@@ -31,7 +34,10 @@ type aggregator struct {
 }
 
 type aggregatorMetrics struct {
-	nbContext int32
+	nbContext      int32
+	nbContextGauge int32
+	nbContextCount int32
+	nbContextSet   int32
 }
 
 func newAggregator(c *Client) *aggregator {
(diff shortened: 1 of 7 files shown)


statsd/aggregator.go
```diff
@@ -115,11 +121,14 @@ func (a *aggregator) flushMetrics() []metric {
 	a.counts = countsMap{}
 	a.countsM.Unlock()
 
-	atomic.AddInt32(&a.nbContext, int32(len(counts)))
 	for _, c := range counts {
 		metrics = append(metrics, c.flushUnsafe())
 	}
 
+	atomic.StoreInt32(&a.nbContextCount, int32(len(counts)))
```
ogaca-dd, line 128: Can `flushMetrics` be called from different threads?

statsd/options.go
```diff
@@ -35,6 +35,8 @@ var (
 	DefaultAggregationFlushInterval = 3 * time.Second
 	// DefaultAggregation
 	DefaultAggregation = false
+	// DevMode
+	DevMode = false
```
ogaca-dd, line 39: Is `DevMode ` used?

ogaca-dd (approved):
LGTM.

statsd/options.go
hush-hush: not yet, good call.

statsd/aggregator.go
hush-hush: Yes, why ?

ogaca-dd: Is the following case possible?
- THREAD1: run `atomic.StoreInt32(&a.nbContextCount, int32(len(counts)))`
- THREAD2: run `atomic.StoreInt32(&a.nbContextCount, int32(len(counts)))`
- THREAD1: read `a.nbContextCount`. The value returned is the one from THREAD2.

hush-hush: Indeed, I moved the `sum` to the telemetry flush

ogaca-dd (approved):
LGTM

PR #170, opened 2020-10-15
Drop sampling rate when aggregation is enabled

hush-hush:
Mixing sampling rate and aggregation doesn't make sens and could be erroneous if users sends the same metrics with different sampling rate.


olivielpeau (approved):
It doesn't hurt to merge this, but we may want to take into account the sample rate even with client-side aggregation in the future

Issue #80, opened 2019-02-12
Use of rand.Float64() reduces parallelism


eliaslevy:
The global default `math/rand` generator is protected by a mutex.  In a highly concurrent application making lots of calls to the DD library and making use of the `rate` parameter to the API sample metrics so as to not drop packets and overload the statsd server, the use of `rand.Float64` in `statsd.Client.send` can become a bottleneck.

Ideally, there would be an alternative API that allows the caller to pass in goroutine local `math.rand.Rand`, which is not protected by a mutex.

arbll:
Hi @eliaslevy,

This looks like it would be a nice improvement. Having an alternative API would work but I am wondering if using a separate `math.rand.Rand` for each client instance and creating one client per goroutine wouldn't be cleaner. 
WDYT ?

eliaslevy:
That would work.  Although it would require folks using the buffered client to reconfigure the timeout or buffer size value to the system to output roughly the same number of packets as when using a single global client.

I went a [somewhat different route](https://github.com/eliaslevy/datadog-go/commit/e3d7304c17a43da70c64e832d627ad4b196b05c6).  I used `sync.Pool` to get a per-CPU RNG.  The only issue is that per my understanding of `sync.Pool`, the RNG is destroyed during GC and recreated again and again.  But it was a quick hack that did not require any API changes.

arbll:
Once https://github.com/DataDog/datadog-go/pull/108 is merged I should be able to simply add a rand per "worker".

matthewdale:
@arbll I opened a PR that uses a `*rand.Rand` per worker to avoid the `math/rand` package-global RNG lock. Please check it out here: https://github.com/DataDog/datadog-go/pull/178

arbll:
Hey - I no longer work on this repo 
cc @hush-hush

hush-hush:
#178 was merged and release (see `v4.3.1`).

PR #178, opened 2020-12-22
Use a worker-specific random source to remove lock contention.

matthewdale:
### What does this PR do ?
Use a separate random number source per worker. Currently, the `worker` function `shouldSample()` calls `rand.Float64()`, which uses a lock to serialize all calls to the package-global pseudorandom number generator, creating lock contention between all instances of `worker`.

Changes:
- Create and seed a separate pseudorandom number generator for every `worker` to remove the lock contention https://github.com/DataDog/datadog-go/issues/80
- Add a test for the `shouldSample()` function to assert that it returns `true` the expected percentage of the time
- Add a benchmark for the `shouldSample()` function to measure improvement from removing lock contention

### Benchmarks against master
```
name            old time/op  new time/op  delta
ShouldSample     444ns ± 0%   190ns ± 1%  -57.26%  (p=0.000 n=8+9)
ShouldSample-2   876ns ± 2%    94ns ± 0%  -89.27%  (p=0.000 n=10+9)
ShouldSample-4  1.17µs ± 1%  0.09µs ± 2%  -92.59%  (p=0.000 n=8+8)
ShouldSample-8  1.25µs ± 3%  0.09µs ± 4%  -92.77%  (p=0.000 n=9+10)
```

### DIFF
--- a/statsd/worker.go
+++ b/statsd/worker.go
@@ -3,12 +3,14 @@ package statsd
 import (
 	"math/rand"
 	"sync"
+	"time"
 )
 
 type worker struct {
 	pool   *bufferPool
 	buffer *statsdBuffer
 	sender *sender
+	random *rand.Rand
 	sync.Mutex
 
 	inputMetrics chan metric
@@ -16,10 +18,16 @@ type worker struct {
 }
 
 func newWorker(pool *bufferPool, sender *sender) *worker {
+	// Note that calling time.Now().UnixNano() repeatedly quickly may return
+	// very similar values. That's fine for seeding the worker-specific random
+	// source because we just need an evenly distributed stream of float values.
+	// Do not use this random source for cryptographic randomness.
+	random := rand.New(rand.NewSource(time.Now().UnixNano()))
 	return &worker{
 		pool:   pool,
 		sender: sender,
 		buffer: pool.borrowBuffer(),
+		random: random,
 		stop:   make(chan struct{}),
 	}
 }
@@ -59,7 +67,7 @@ func (w *worker) processMetric(m metric) error {
 }
 
 func (w *worker) shouldSample(rate float64) bool {
-	if rate < 1 && rand.Float64() > rate {
+	if rate < 1 && w.random.Float64() > rate {
 		return false
 	}
 	return true
--- a/statsd/worker.go
+++ b/statsd/worker.go
@@ -18,7 +18,11 @@ type worker struct {
 }
 
 func newWorker(pool *bufferPool, sender *sender) *worker {
-	// Note that calling time.Now().UnixNano() repeatedly quickly may return
+	// Each worker uses its own random source to prevent workers in separate
+	// goroutines from contending for the lock on the "math/rand" package-global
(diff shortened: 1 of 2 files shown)


statsd/worker_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestShouldSample(t *testing.T) {
+	rates := []float64{0.01, 0.05, 0.1, 0.25, 0.5, 0.75, 0.9, 0.99, 1.0}
+	iterations := 50_000
+	worker := newWorker(newBufferPool(1, 1, 1), nil)
+	for _, rate := range rates {
+		rate := rate // Capture range variable.
```
hush-hush, line 15: I don't think we need this since `t.Run` will block until the  test is done.

statsd/worker.go
```diff
@@ -3,23 +3,31 @@ package statsd
...
 	inputMetrics chan metric
 	stop         chan struct{}
 }
 
 func newWorker(pool *bufferPool, sender *sender) *worker {
+	// Note that calling time.Now().UnixNano() repeatedly quickly may return
+	// very similar values. That's fine for seeding the worker-specific random
+	// source because we just need an evenly distributed stream of float values.
+	// Do not use this random source for cryptographic randomness.
+	random := rand.New(rand.NewSource(time.Now().UnixNano()))
```
hush-hush, line 25: Do you mind adding a comment quickly explaining why each worker has its own random ?

hush-hush (changes requested):
Thanks for the PR, this looks great. I added a few nit-pick but otherwise this looks ready to go.

statsd/worker.go
matthewdale: Sure, will add a comment.

statsd/worker_test.go
matthewdale: Ah, you're correct, that makes me realize I'd intended to add `t.Parallel()` to run each subtest in parallel. I'll make that change and push it.

hush-hush (approved):
Thanks for the PR !

statsd/worker.go
```diff
@@ -59,7 +71,7 @@ func (w *worker) processMetric(m metric) error {
 }
 
 func (w *worker) shouldSample(rate float64) bool {
-	if rate < 1 && rand.Float64() > rate {
+	if rate < 1 && w.random.Float64() > rate {
```
rafaeljusto, line 74: @matthewdale @hush-hush Shouldn't we have a local lock here? From the [docs](https://golang.org/pkg/math/rand/):
> The default Source is safe for concurrent use by multiple goroutines, **but Sources created by NewSource are not.**

I think this is causing race conditions.

matthewdale: You're correct. When `MutexMode` is enabled, the call to `worker.processMetric()` is made from the same goroutine as the call to the top-level metric emit function (e.g. `Count`). I was previously under the impression that all calls to `worker.processMetric()` would be made from a single goroutine per worker, but that's not the case.

I'm working on a fix for the data race bug and will submit it when it's ready.

PR #179, opened 2021-01-27
Fix worker random race condition

hush-hush:
`rand.NewSource` is not thread safe. Each worker needs its own random and random-lock.

### DIFF
--- a/statsd/worker.go
+++ b/statsd/worker.go
@@ -7,10 +7,11 @@ import (
 )
 
 type worker struct {
-	pool   *bufferPool
-	buffer *statsdBuffer
-	sender *sender
-	random *rand.Rand
+	pool       *bufferPool
+	buffer     *statsdBuffer
+	sender     *sender
+	random     *rand.Rand
+	randomLock sync.Mutex
 	sync.Mutex
 
 	inputMetrics chan metric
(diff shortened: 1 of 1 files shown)


statsd/worker.go
```diff
@@ -71,6 +73,9 @@ func (w *worker) processMetric(m metric) error {
 }
 
 func (w *worker) shouldSample(rate float64) bool {
+	// rand.NewSource is not thread safe
+	w.randomLock.Lock()
+	defer w.randomLock.Unlock()
```
olivielpeau, line 78: `defer` had an overhead until Go 1.14, since we support older go versions we may want to avoid using it here since it's a hot path

olivielpeau (approved):
Not a blocker, I'm just curious: what's the impact of this change on the benchmark of the original PR: https://github.com/DataDog/datadog-go/pull/178 ?

statsd/worker.go
```diff
@@ -71,9 +73,15 @@ func (w *worker) processMetric(m metric) error {
...
 func (w *worker) shouldSample(rate float64) bool {
+	// rand.NewSource is not thread safe.
+	// TODO: use defer once the lowest Go version we support is 1.14 (defer
+	// has an overhead before that).
+	w.randomLock.Lock()
 	if rate < 1 && w.random.Float64() > rate {
+		w.randomLock.Unlock()
 		return false
 	}
+	w.randomLock.Unlock()
```
ogaca-dd, line 84: Just a note: It can be rewrite as
```go
w.randomLock.Lock()
defer w.randomLock.Unlock()
return rate >= 1 || w.random.Float64() <= rate
```

ogaca-dd (approved):
LGTM

statsd/worker.go
olivielpeau: it's better to avoid `defer`, see my earlier comment: https://github.com/DataDog/datadog-go/pull/179#discussion_r565463981

statsd/worker.go
truthbk: details: https://medium.com/i0exception/runtime-overhead-of-using-defer-in-go-7140d5c40e32

statsd/worker.go
```diff
@@ -71,9 +73,15 @@ func (w *worker) processMetric(m metric) error {
 }
 
 func (w *worker) shouldSample(rate float64) bool {
+	// rand.NewSource is not thread safe.
```
truthbk, line 76: Nit: for clarity, I suggest:
```suggestion
	// sources created by rand.NewSource() (ie. w.random) are not thread safe.
```

truthbk (approved):

statsd/worker.go
hush-hush: See comment from olivier about defer overhead

PR #180, opened 2021-01-29
Add a Client test that can detect data races when emitting metrics from separate goroutines.

matthewdale:
### What does this PR do?
Adds a `statsd.Client` test that can detect data races when recording metrics from separate goroutines.

Inspired by the bug introduced in https://github.com/DataDog/datadog-go/pull/178 and fixed with https://github.com/DataDog/datadog-go/commit/5cfbb1464e67fe05d7a6ba65c1a4c7dbe06b8420.

### DIFF
--- a/statsd/statsd_test.go
+++ b/statsd/statsd_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


matthewdale:
@hush-hush thanks for fixing the data race bug! Here's the test I used to confirm it was indeed an issue. Hopefully this will help catch data races in the future.

hush-hush (approved):

hush-hush:
Thanks for the test @matthewdale !

PR #182, opened 2021-02-08
statsd: add support for Windows Pipes

gbbr:
This change adds support for Windows Named Pipes. They may be used when
this option is [enabled in the Datadog Agent](https://github.com/DataDog/datadog-agent/pull/6830).

### DIFF
--- a/statsd/pipe.go
+++ b/statsd/pipe.go
@@ -1,0 +1,9 @@
+// +build !windows
+
+package statsd
+
+import "errors"
+
+func newWindowsPipeWriter(pipepath string) (statsdWriter, error) {
+	return nil, errors.New("Windows Named Pipes are not supported on other operating systems than Windows")
+}
--- a/statsd/pipe.go
+++ b/statsd/pipe.go
@@ -5,5 +5,5 @@ package statsd
 import "errors"
 
 func newWindowsPipeWriter(pipepath string) (statsdWriter, error) {
-	return nil, errors.New("Windows Named Pipes are not supported on other operating systems than Windows")
+	return nil, errors.New("Windows Named Pipes are only supported on Windows")
 }
--- a/statsd/pipe_windows.go
+++ b/statsd/pipe_windows.go
@@ -1,0 +1,22 @@
+// +build windows
+
+package statsd
+
+import (
+	"errors"
+	"net"
+	"time"
+
+	"github.com/Microsoft/go-winio"
+)
+
+type pipeWriter struct{ net.Conn }
+
+func (pipeWriter) SetWriteTimeout(_ time.Duration) error {
+	return errors.New("SetWriteTimeout is not supported on Windows Named Pipes")
+}
+
+func newWindowsPipeWriter(pipepath string) (*pipeWriter, error) {
+	conn, err := winio.DialPipe(pipepath, nil)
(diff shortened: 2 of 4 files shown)


statsd/statsd.go
```diff
@@ -55,6 +55,12 @@ traffic instead of UDP.
 */
 const UnixAddressPrefix = "unix://"
 
+/*
+WindowsPipeAddressPrefix holds the prefix to use to enable Windows Named Pipes
+traffic instead of UDP.
+*/
+const WindowsPipeAddressPrefix = `\\.\pipe\`
```
hush-hush, line 62: Does this need to be public ?

statsd/statsd.go
```diff
@@ -94,8 +100,9 @@ const (
 )
 
 const (
-	WriterNameUDP string = "udp"
-	WriterNameUDS string = "uds"
+	WriterNameUDP     string = "udp"
+	WriterNameUDS     string = "uds"
+	WriterWindowsPipe string = "pipe"
```
hush-hush, line 105: Same about those, could we make them private

hush-hush (changes requested):
Could we also update the comment line 370 of statsd/statsd.go ?

Other than this an the few comments it looks ready to go.

statsd/pipe.go
```diff
@@ -0,0 +1,9 @@
+// +build !windows
+
+package statsd
+
+import "errors"
+
+func newWindowsPipeWriter(pipepath string) (statsdWriter, error) {
+	return nil, errors.New("Windows Named Pipes are not supported on other operating systems than Windows")
```
lucaspimentel, line 8: Consider rewording this for clarity.
```suggestion
	return nil, errors.New("Windows Named Pipes are only supported on Windows")
```

statsd/statsd.go
gbbr: It absolutely does not, but then again neither do the UDS ones. I was just following what's there.

statsd/statsd.go
gbbr: We could, but that would be a breaking change. Technically it's not ok to do that in open-source projects but if you want me too I can do it.

hush-hush: Indeed missed that. Let's keep those public.

statsd/statsd.go
hush-hush: We have too many public method/attribute in this package. The next major is going to make everything that doesn't need to be public private.

I guess we could keep it public for consistency for now since it's a const.

hush-hush (approved):

statsd/pipe_windows.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+type pipeWriter struct{ net.Conn }
+
+func (pipeWriter) SetWriteTimeout(_ time.Duration) error {
+	return errors.New("SetWriteTimeout is not supported on Windows Named Pipes")
+}
+
+func newWindowsPipeWriter(pipepath string) (*pipeWriter, error) {
+	conn, err := winio.DialPipe(pipepath, nil)
```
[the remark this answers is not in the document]
gbbr: Resolved in https://github.com/DataDog/datadog-go/pull/185

PR #185, opened 2021-02-11
statsd: ensure Windows Pipe writer tries to reconnect when losing connection

gbbr:
This changes ensures that the Windows Pipe statsdWriter attempts
reconnecting after losing a connection.

### DIFF
--- a/statsd/pipe_windows.go
+++ b/statsd/pipe_windows.go
@@ -3,20 +3,80 @@
 package statsd
 
 import (
-	"errors"
 	"net"
+	"sync"
 	"time"
 
 	"github.com/Microsoft/go-winio"
 )
 
-type pipeWriter struct{ net.Conn }
+const defaultPipeTimeout = 1 * time.Millisecond
 
-func (pipeWriter) SetWriteTimeout(_ time.Duration) error {
-	return errors.New("SetWriteTimeout is not supported on Windows Named Pipes")
+type pipeWriter struct {
+	mu       sync.RWMutex
+	conn     net.Conn
+	timeout  time.Duration
+	pipepath string
+}
+
+func (p *pipeWriter) SetWriteTimeout(d time.Duration) error {
+	p.timeout = d
+	return nil
+}
+
+func (p *pipeWriter) Write(data []byte) (n int, err error) {
+	conn, err := p.ensureConnection()
+	if err != nil {
+		return 0, err
+	}
+
+	p.mu.RLock()
+	conn.SetWriteDeadline(time.Now().Add(p.timeout))
+	p.mu.RUnlock()
+
+	n, err = conn.Write(data)
+	if err != nil {
+		if e, ok := err.(net.Error); !ok || !e.Temporary() {
+			// disconnected; retry again on next attempt
+			p.mu.Lock()
+			p.conn = nil
+			p.mu.Unlock()
+		}
+	}
+	return n, err
+}
+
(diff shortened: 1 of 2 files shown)


hush-hush (approved):

statsd/pipe_windows.go
```diff
@@ -3,20 +3,82 @@
...
+	conn, err := p.ensureConnection()
+	if err != nil {
+		return 0, err
+	}
+
+	p.mu.RLock()
+	conn.SetWriteDeadline(time.Now().Add(p.timeout))
+	p.mu.RUnlock()
+
+	n, err = conn.Write(data)
```
ogaca-dd, line 39: What happens if 2 go routines call `conn.Write(data)` at the same time. Is the server able to read the first data and then the second data or the data from call 1 and 2 are mixed?

statsd/pipe_windows.go
```diff
@@ -3,20 +3,82 @@
...
+	if err != nil {
+		if e, ok := err.(net.Error); !ok || !e.Temporary() {
+			// disconnected; retry again on next attempt
```
ogaca-dd, line 42: In case of disconnection, what do you think about retrying the sending of `data`?

statsd/pipe_windows_test.go
```diff
@@ -4,17 +4,34 @@ package statsd
...
 	"os"
 	"testing"
+	"time"
 
 	"github.com/Microsoft/go-winio"
 )
 
+// acceptOne accepts one single connection from ln, reads 512 bytes from it
+// and sends it to the out channel, afterwards closing the connection.
+func acceptOne(t *testing.T, ln net.Listener, out chan string) {
```
ogaca-dd, line 17: 👍

statsd/pipe_windows_test.go
```diff
@@ -23,32 +40,83 @@ func TestPipeWriter(t *testing.T) {
...
 		t.Fatalf("Expected %q, got %q", exp, got)
 	}
 }
+
+func TestPipeWriterReconnect(t *testing.T) {
+	f, err := ioutil.TempFile("", "test-pipe-")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.Remove(f.Name())
```
ogaca-dd, line 66: The pipe name can be any string and does not require a file to exist unlike to Unix domain socket.

ogaca-dd (commented):
Great PR, just a few comments.

statsd/pipe_windows.go
gbbr: I don't know. The same question applies to the UDS implementation I guess. The only way out of it is introducing a mutex. Do we want that? I'd think no.

gbbr (commented):
Thanks Olivier! Let's keep it consistent with the UDS implementation, as discussed on Slack.

statsd/pipe_windows.go
ogaca-dd: As far as I remember, UDS does not have this issue as it uses datagram.

gbbr: I wrote a test to try this out. It creates 100 goroutines each sending 100 metrics simultaneously and then checks the output confirming that each individual metric is there. It seems fine, even when running the tests many times in a row (with and without `-race` flag and with `-count=20`). Code is here: https://github.com/DataDog/datadog-go/compare/gbbr/test-concur?expand=1

PR #187, opened 2021-03-01
Implement ChannelMode and sampling rate for extended aggregation

hush-hush:
When using extended aggregation we still want to respect sampling rate for histograms, distribution and timing as it will have direct consequences on the Agent (number of metrics to aggregate).

For app sending a very high number of metrics the aggregator implements ChannelMode to avoid lock contention when generating 
 andom numbers.

### DIFF
--- a/statsd/aggregator.go
+++ b/statsd/aggregator.go
@@ -11,58 +11,8 @@ type (
 	countsMap         map[string]*countMetric
 	gaugesMap         map[string]*gaugeMetric
 	setsMap           map[string]*setMetric
-	bufferedMetricMap map[string]*histogramMetric
+	bufferedMetricMap map[string]*bufferedMetric
 )
-
-// bufferedMetricContexts represent the contexts for Histograms, Distributions
-// and Timing. Since those 3 metric types behave the same way and are sampled
-// with the same type they're represented by the same class.
-type bufferedMetricContexts struct {
-	nbContext int32
-	mutex     sync.RWMutex
-	values    bufferedMetricMap
-	newMetric func(string, float64, string) *bufferedMetric
-}
-
-func newBufferedContexts(newMetric func(string, float64, string) *bufferedMetric) bufferedMetricContexts {
-	return bufferedMetricContexts{
-		values:    bufferedMetricMap{},
-		newMetric: newMetric,
-	}
-}
-
-func (bc *bufferedMetricContexts) flush(metrics []metric) []metric {
-	bc.mutex.Lock()
-	values := bc.values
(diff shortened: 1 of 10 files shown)


statsd/buffered_metric_context.go
```diff
@@ -0,0 +1,75 @@
...
+
+	for _, d := range values {
+		metrics = append(metrics, d.flushUnsafe())
+	}
+	atomic.AddInt32(&bc.nbContext, int32(len(values)))
+	return metrics
+}
+
+func (bc *bufferedMetricContexts) sample(name string, value float64, tags []string, rate float64) error {
+	if !shouldSample(rate, bc.random, &bc.randomLock) {
```
truthbk, line 54: I know there's no easy way around this, I'm a little concerned about all this locking here. We have one lock per metric type which is still a pretty coarse granularity. Can't really think of a good workaround.

statsd/aggregator.go
```diff
@@ -11,59 +11,9 @@ type (
 	countsMap         map[string]*countMetric
 	gaugesMap         map[string]*gaugeMetric
 	setsMap           map[string]*setMetric
-	bufferedMetricMap map[string]*histogramMetric
+	bufferedMetricMap map[string]*bufferedMetric
```
truthbk, line 14: So, I'm wondering if we should consider `sync.Map`? It's not totally clear to me, but we _might_ get better performance. On paper a `sync.Map` might offer better performance when used as follows:

```
(1) when the entry for a given key is only ever written once but read many times, as in caches that only grow,

(2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys.
```

I think we have a bit of (1) in that we create the context once and then sample over and over, but it might not fit the description fully. We don't have super high concurrency for (2), and the sets of keys would not be disjoint, so I'm not sure that apples. In any case, might be worth giving it thought to decide if we should keep the current implementation.

truthbk (approved):
Super clean, I really don't have much to say. Added a couple of comments regarding areas that may be a bit of a concern in terms of potential performance hurdles - difficult to overcome in any case - but that we could maybe look into. 

Feel free to merge.

statsd/buffered_metric_context.go
hush-hush: That's true. Also it's the same overhead as the default configuration right now. When using the default setting (ie mutext mode), we have the same lock and random. So the overhead should be the same with and without extended aggregation.

statsd/aggregator.go
hush-hush: That's a really good point. I'll do that in a different PR !

PR #189, opened 2021-03-12
Detect address from env before resolving it

hoshsadiq:
Seems the address is resolved from the environment after it is determined that we're not using unix socket/windows pipe. In our case, the application passes in an empty address, as we're relying on the client to determine the address from the environment. However, if `DD_AGENT_HOST` is set and prefixed with `unix://` or `\\.\pipe\`, it won't have detected this and will eventually just falls back to `UDPWriter` (in this case, it also always adds the default port). This should fix that by determining retrieving the address from the environment before that logic kicks in.


hush-hush:
Thanks a lot for the PR @hoshsadiq. I opened #192 to answer the issue and re-organize the code and tests.

hush-hush:
This has been fixed in `v4.5.1`. Thanks again for bringing it up.

PR #190, opened 2021-03-15
[win pipes] Defer connection establishment to first write

olivielpeau:
This allows the client creation call (`New`) with a windows named pipe
target address to be successful even if the Agent hasn't created the
named pipe yet.

This new behavior is consistent with the UDS writer.

I haven't tested the change on Windows, relying on CI to catch any regression on Windows.


hush-hush (approved):

gbbr (approved):
👍 Thanks Olivier

PR #197, opened 2021-05-14
Add go.mod and go.sum files to project

### LINKED ISSUE #119: Go Modules Support

danopia:
Fixes #119

Closes #150 - that PR has been stale with changes requested for more than a quarter now.

In `go.mod` I've specified `go 1.13` which is the oldest version which I see CI configured for in this repository. From what I can tell, that directive should indicate the "target level" of the project. I don't see a strong reason to set it to any particular value given that 1.13, 1.14, and 1.15 are all tested by CI and should all work.
z
The steps to make this file are quite reproducible:

```sh
go mod init github.com/DataDog/datadog-go
go mod tidy
# ... and also update go version in go.mod
```

Given that a standard go toolchain is used going forward, both of this files should be automatically updated as the project evolves. However this might not happen if the `GO111MODULE=off` variable is generally used by contributors to opt-out of these modules.

The build failure seems unrelated; I don't know how to rebuild other than a force push.


hush-hush:
Thanks for the PR @danopia,

This is planned for the next major of datadog-go which, I hope, should come in the next quarter.

danopia:
Thanks for the info.

I opened this fresh PR in hopes it could merge through without any blocker. If the gomod support is being held until next release, and is already planned, perhaps I should close this and let a fresh go.sum get written up when the team is prepping for the release? As long as the files get added somehow :)

hush-hush:
HI @danopia ,

The new major version was delayed more than we expected but it's finally released. We've added go modules support.

I'm closing this PR.

PR #198, opened 2021-05-14
Added EOL termination

gh123man:
Messages should be '\n' terminated always, using the EOL delimiter helps us understand if a metric has been truncated at the end of a payload. The server side will allow enabling a mode to drop metrics that are not properly EOL terminated to avoid processing partial metrics.

Same as change made for java DSD client https://github.com/DataDog/java-dogstatsd-client/pull/130


truthbk (approved):
This looks good to me! Sorry about all those one char changes to all the tests! 👍

hush-hush (approved):

PR #199, opened 2021-05-28
DefaultAggregationFlushInterval: 3s -> 2s; reduce sampling aliasing

evanj:
When sending a count metric that increments at a constant 1000/sec,
client-side aggregation causes it to appear noisy, reporting values
of 900, 900, 1200 instead of the constant 1000/sec. This appears to
be mostly caused by mismatched reporting intervals. The client-side
aggregation is reporting every 3 seconds, while the agent is
reporting every 10 seconds. This means in each agent bucket, the
values are: 9000, 9000, 12000, ... . Setting the flush interval to
2 seconds corrects this issue.


evanj:
I spent some time trying to figure out if I could add a test for this, but I'm not sure it is worth the effort. I'm happy to try if you have suggestions for how to do it.

hush-hush (approved):

PR #205, opened 2021-07-01
Fix telemetry using the global namespace

hush-hush:
The telemetry metrics should not use the user namespace.
Before this PR: the metric namespace would still be added to telemetry metrics when telemetry is not configured with its own endpoint.

This commit also include a rework in the tests that allows us to test the telemetry from start to finish (end-to-end)

### DIFF
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -497,9 +497,6 @@ func (c *Client) FlushTelemetryMetrics() ClientMetrics {
 }
 
 func (c *Client) send(m metric) error {
-	m.globalTags = c.Tags
-	m.namespace = c.Namespace
-
 	h := hashString32(m.name)
 	worker := c.workers[h%uint32(len(c.workers))]
 
@@ -545,7 +542,7 @@ func (c *Client) Gauge(name string, value float64, tags []string, rate float64) 
 	if c.agg != nil {
 		return c.agg.gauge(name, value, tags)
 	}
-	return c.send(metric{metricType: gauge, name: name, fvalue: value, tags: tags, rate: rate})
+	return c.send(metric{metricType: gauge, name: name, fvalue: value, tags: tags, rate: rate, globalTags: c.Tags, namespace: c.Namespace})
 }
 
 // Count tracks how many times something happened per second.
@@ -557,7 +554,7 @@ func (c *Client) Count(name string, value int64, tags []string, rate float64) er
 	if c.agg != nil {
 		return c.agg.count(name, value, tags)
 	}
-	return c.send(metric{metricType: count, name: name, ivalue: value, tags: tags, rate: rate})
+	return c.send(metric{metricType: count, name: name, ivalue: value, tags: tags, rate: rate, globalTags: c.Tags, namespace: c.Namespace})
 }
 
 // Histogram tracks the statistical distribution of a set of values on each host.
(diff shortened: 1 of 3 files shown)


statsd/test_helpers_test.go
```diff
@@ -0,0 +1,411 @@
...
+
+	aggregationEnabled  = true
+	aggregationDisabled = false
+
+	extendedAggregationEnabled  = true
+	extendedAggregationDisabled = false
+
+	telemetryEnabled  = true
+	telemetryDisabled = false
+)
```
olivielpeau, line 79: what are these constants for? (they seem to be unused)

statsd/test_helpers_test.go
```diff
@@ -0,0 +1,411 @@
...
+	finalTags := ts.getFinalTags()
+
+	return append(
+		res,
+		"_e{5,5}:hello|world"+finalTags,
+		"_sc|hello|1"+finalTags,
+	)
+}
+
+func (ts *testServer) sendBasicAggregationMetrics(client *Client) []string {
```
olivielpeau, line 376: seems to be unused

statsd/test_helpers_test.go
```diff
@@ -0,0 +1,411 @@
...
+
+	finalTags := ts.getFinalTags(tags...)
+	return []string{
+		ts.namespace + "set:my_id|s" + finalTags,
+		ts.namespace + "gauge:21|g" + finalTags,
+		ts.namespace + "count:4|c" + finalTags,
+	}
+}
+
+func (ts *testServer) sendExtendedBasicAggregationMetrics(client *Client) []string {
```
olivielpeau, line 393: ditto

statsd/statsd.go
```diff
@@ -497,9 +497,6 @@ func (c *Client) FlushTelemetryMetrics() ClientMetrics {
 }
 
 func (c *Client) send(m metric) error {
-	m.globalTags = c.Tags
```
olivielpeau, line 500: I think removing the global tags from telemetry metrics makes sense, but note that in the future we may want to allow the telemetry metrics to have global tags as well (this can help differentiate the telemetry metrics coming from 2 different clients with the same version running on the same host). Let's follow up in chat.

olivielpeau (commented):
Left a few comments, but overall LGTM.

If I understand correctly (could you update the PR description if this looks correct):
- before this PR: the metric namespace (and global tags) would still be added to telemetry metrics when telemetry is not configured with its own endpoint.
- after this PR: the metric namespace (and global tags) are _not_ added to telemetry metrics in all cases, which is the correct behavior.

statsd/test_helpers_test.go
hush-hush: Nowhere anymore :+1:

statsd/test_helpers_test.go
hush-hush: True, but it will when we use the new test helper everywhere. We have a lot of duplicate code in our test that the new helpers would remove. I didn't want this PR to get out of hand. But I plan an upgrading the test suite with the new test helper soon.

statsd/test_helpers_test.go
hush-hush: Same here

olivielpeau (approved):
LGTM!

(and to correct my previous comment: global tags were already sent with telemetry metrics before this PR, and will continue to be with this PR).

PR #206, opened 2021-07-01
Improve test by using the new test helper

hush-hush:
Reduce code duplication and now test what is produce by the client rather what it's internal state.


statsd/statsd.go
```diff
@@ -332,11 +333,15 @@ func newWithWriter(w statsdWriter, o *Options, writerName string) (*Client, erro
...
 	// Inject values of DD_* environment variables as global tags.
+	extraTags := []string{}
 	for envName, tagName := range ddEnvTagsMapping {
 		if value := os.Getenv(envName); value != "" {
-			c.Tags = append(c.Tags, fmt.Sprintf("%s:%s", tagName, value))
+			extraTags = append(extraTags, fmt.Sprintf("%s:%s", tagName, value))
 		}
 	}
+	// We sort env tags to ease testing since map order access is not guarantee
+	sort.Strings(extraTags)
```
vickenty, line 343: Is this necessary? The tests seem to sort tags anyway. I'm not sure how perf-critical this code is, but if we do need tags in specific order, perhaps we could convert `ddEnvTagsMapping` to be a slice instead. Seems unnecessary to keep sorting what is basically a constant.

statsd/statsd.go
```diff
@@ -70,17 +70,13 @@ const (
...
-	"DD_ENV": "env",
-	// The name of the running service.
-	"DD_SERVICE": "service",
-	// The current version of the running service.
-	"DD_VERSION": "version",
+var ddEnvTagsMapping = [][]string{
+	[]string{"DD_ENTITY_ID", "dd.internal.entity_id"}, // Client-side entity ID injection for container tagging.
+	[]string{"DD_ENV", "env"},                         // The name of the env in which the service runs.
+	[]string{"DD_SERVICE", "service"},                 // The name of the running service.
+	[]string{"DD_VERSION", "version"},                 // The current version of the running service.
```
vickenty, line 79:
```suggestion
var ddEnvTagsMapping = []struct{envName, tagName}{
	{"DD_ENTITY_ID", "dd.internal.entity_id"}, // Client-side entity ID injection for container tagging.
	{"DD_ENV", "env"},                         // The name of the env in which the service runs.
	{"DD_SERVICE", "service"},                 // The name of the running service.
	{"DD_VERSION", "version"},                 // The current version of the running service.
```

And maybe const too?

PR #207, opened 2021-07-05
Move CI to circleCI

hush-hush:


.circleci/config.yml
```diff
@@ -0,0 +1,122 @@
...
+              $SUDO install --owner=${USER} -d /usr/local/go
+            fi
+
+            echo "Installing the requested version of Go."
+
+            curl --fail --location -sS "https://dl.google.com/go/go<<parameters.version >>.linux-<< parameters.arch >>.tar.gz" \
+            | sudo tar --no-same-owner --strip-components=1 --gunzip -x -C /usr/local/go/
+
+            echo "export PATH=$PATH:/usr/local/go/bin" >> $BASH_ENV
+            $SUDO chown -R $(whoami): /usr/local/go
```
sgnn7, line 48: Nit: Would be nice to have all params with vars enclosed in quotes to prevent unwanted token splitting. I know we control most of them but it's good practice to use on shell scripts.

sgnn7 (approved):
@hush-hush One tiny nit but LGTM and 👍 on the change.

PR #213, opened 2021-08-04
Removing 'SetWriteTimeout' method in favor of WithWriteTimeout

hush-hush:
WithWriteTimeoutUDS has been removed in favor of WithWriteTimeout (since we now have UDS and named pipe that support write timeout).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 ## Breaking changes
 
 - `WithDevMode` option has been removed. The extended telemetry it enabled is now part of the default telemetry.
+- `WithWriteTimeoutUDS` option has been renamed `WithWriteTimeout` since it also impact named pipe transport.
+- `SetWriteTimeout` method has been removed in favor of `WithWriteTimeout` option.
 
 # 4.8.1 / 2021-07-09
 
--- a/statsd/options.go
+++ b/statsd/options.go
@@ -24,8 +24,8 @@ var (
 	DefaultBufferShardCount = 32
 	// DefaultSenderQueueSize is the default value for the DefaultSenderQueueSize option
 	DefaultSenderQueueSize = 0
-	// DefaultWriteTimeoutUDS is the default value for the WriteTimeoutUDS option
-	DefaultWriteTimeoutUDS = 100 * time.Millisecond
+	// DefaultWriteTimeout is the default value for the WriteTimeout option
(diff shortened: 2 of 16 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 ## Breaking changes
 
 - 'WithDevMode' option has been removed. The extended telemetry it enabled is now part of the default telemetry.
+- 'WithWriteTimeoutUDS' option has been renamed `WithWriteTimeout` since it also impact named pipe transport.
+- 'SetWriteTimeout' method has been removed in favor of 'WithWriteTimeout' option.
```
sgnn7, line 11: non-blocking nit:
```suggestion
- `WithWriteTimeoutUDS` option has been renamed `WithWriteTimeout` since it also impact named pipe transport.
- `SetWriteTimeout` method has been removed in favor of `WithWriteTimeout` option.
```

statsd/options.go
```diff
@@ -69,8 +69,8 @@ type Options struct {
 	// The magic value 0 will set the option to the optimal size for the transport
 	// protocol used when creating the client: 2048 for UDP and 512 for UDS.
 	SenderQueueSize int
-	// WriteTimeoutUDS is the timeout after which a UDS packet is dropped.
-	WriteTimeoutUDS time.Duration
+	// WriteTimeout is the timeout after which a UDS packet is dropped.
```
sgnn7, line 72: nit: is this for both UDS and UDP? If so, we should remove the specificity.
```suggestion
	// WriteTimeout is the timeout after which a packet is dropped.
```

statsd/options.go
```diff
@@ -212,10 +212,10 @@ func WithSenderQueueSize(senderQueueSize int) Option {
 	}
 }
 
-// WithWriteTimeoutUDS sets the WriteTimeoutUDS option.
-func WithWriteTimeoutUDS(writeTimeoutUDS time.Duration) Option {
+// WithWriteTimeout sets the WriteTimeout option.
```
sgnn7, line 215: nit: This comment could be a bit better but I know it's there from the old code

sgnn7 (approved):
A few nits but LGTM

statsd/options.go
hush-hush: I have another PR rewriting the doc

PR #215, opened 2021-08-04
Remove 'Namespace' and 'Tags' from the public API.

hush-hush:
Both attribute were not thread safe. 'WithNamespace' and WithTags'
options should be used instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
   + `WriterNameUDS`
   + `WriterWindowsPipe`
   + `TelemetryInterval`
+- Field `Client.Namespace` is now private, please use the `WithNamespace` option.
+- Field `Client.Tags` is now private, please use the `WithTags` option.
 
 # 4.8.1 / 2021-07-09
 
--- a/statsd/client_test.go
+++ b/statsd/client_test.go
@@ -1,6 +1,6 @@
 // Copyright 2013 Ooyala, Inc.
 
-package statsd_test
+package statsd
 
 import (
 	"io"
@@ -12,7 +12,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/DataDog/datadog-go/statsd"
 	"github.com/stretchr/testify/assert"
 )
 
@@ -54,16 +53,12 @@ func TestClientUDP(t *testing.T) {
 	}
 	defer server.Close()
 
-	client, err := statsd.New(addr)
+	client, err := New(addr)
 	if err != nil {
 		t.Fatal(err)
 	}
 
 	clientTest(t, server, client)
-}
-
-type statsdWriterWrapper struct {
-	io.WriteCloser
 }
 
 func TestClientWithConn(t *testing.T) {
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,8 @@
   + 'WriterNameUDS'
   + 'WriterWindowsPipe'
   + 'TelemetryInterval'
+- Field 'Client.Namespace' is now private, please use the 'WithNamespace' option.
+- Field 'Client.Tags' is now private, please use the 'WithTags' option.
```
sgnn7, line 18: nit
```suggestion
- Field `Client.Namespace` is now private, please use the `WithNamespace` option.
- Field `Client.Tags` is now private, please use the `WithTags` option.
```

statsd/client_test.go
```diff
@@ -54,36 +53,32 @@ func TestClientUDP(t *testing.T) {
...
 	clientTest(t, server, client)
 }
 
-func clientTest(t *testing.T, server io.Reader, client *statsd.Client) {
+func clientTest(t *testing.T, server io.Reader, client *Client) {
 	for _, tt := range dogstatsdTests {
-		client.Namespace = tt.GlobalNamespace
-		client.Tags = tt.GlobalTags
+		client.namespace = tt.GlobalNamespace
+		client.tags = tt.GlobalTags
```
sgnn7, line 81: I know these are tests but I'm not sure that we should be trying to use non-public fields here otherwise we're not really testing the same thing that the user will be interacting with.Is there a way to run these with the public side of the APIs?

statsd/client_test.go
```diff
@@ -121,13 +116,13 @@ func TestBufferedClient(t *testing.T) {
...
-	client, err := statsd.NewBuffered(addr, bufferLength)
+	client, err := NewBuffered(addr, bufferLength)
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	client.Namespace = "foo."
-	client.Tags = []string{"dd:2"}
+	client.namespace = "foo."
+	client.tags = []string{"dd:2"}
```
sgnn7, line 125: Ditto on the use of private fields

statsd/uds_test.go
```diff
@@ -98,8 +73,8 @@ func (suite *UdsTestSuite) TestClientUDS() {
 	}
 
 	for _, tt := range dogstatsdTests {
-		client.Namespace = tt.GlobalNamespace
-		client.Tags = tt.GlobalTags
+		client.namespace = tt.GlobalNamespace
+		client.tags = tt.GlobalTags
```
sgnn7, line 77: Ditto on use of private fields

sgnn7 (commented):
@hush-hush Left a comment but the rest lgtm

statsd/client_test.go
hush-hush: Very good point. The tests are a bit of a mess right now. It's one of the main improvement I want to bring in V5. We already included some test helpers that will ease the testing issue and will allow us to rewrite those test in a better way (and remove duplicated tests).

It will be done is a future PRs

PR #224, opened 2021-08-25
Add the method 'GetTelemetry' to retrieve the internal telemetry

hush-hush:
This allows client app to dynamically adapt its behavior (sampling rate, metrics
submission patterns, number of metric sent ...) based on packet drop for
example.

Metrics are now forever increasing making it resilient to packet drop. This means they are not sent as gauge.

### DIFF
--- a/statsd/aggregator.go
+++ b/statsd/aggregator.go
@@ -15,9 +15,9 @@ type (
 )
 
 type aggregator struct {
-	nbContextGauge int32
-	nbContextCount int32
-	nbContextSet   int32
+	nbContextGauge uint64
+	nbContextCount uint64
+	nbContextSet   uint64
 
 	countsM sync.RWMutex
 	gaugesM sync.RWMutex
@@ -41,16 +41,6 @@ type aggregator struct {
 	inputMetrics    chan metric
 	stopChannelMode chan struct{}
 	wg              sync.WaitGroup
-}
-
-type aggregatorMetrics struct {
-	nbContext             int32
-	nbContextGauge        int32
-	nbContextCount        int32
-	nbContextSet          int32
-	nbContextHistogram    int32
-	nbContextDistribution int32
-	nbContextTiming       int32
 }
 
 func newAggregator(c *Client) *aggregator {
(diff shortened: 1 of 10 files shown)


statsd/sender.go
```diff
@@ -5,23 +5,21 @@ import (
...
-	TotalSentPayloads             uint64
-	TotalDroppedPayloads          uint64
-	TotalDroppedBytes             uint64
-	TotalDroppedPayloadsQueueFull uint64
-	TotalDroppedBytesQueueFull    uint64
-	TotalDroppedPayloadsWriter    uint64
-	TotalDroppedBytesWriter       uint64
+// senderTelemetry contains telemetry about the health of the sender
+type senderTelemetry struct {
+	totalPayloadsSent             uint32
```
truthbk, line 10: Why not `uint64`? Now that counters are forever increasing there's an even greater likelihood of overflow (`uint32` max value is 4294967294, which may not be large enough for long-running applications), we could use the additional bits. Am I missing something?

statsd/statsd.go
```diff
@@ -210,18 +210,18 @@ type Client struct {
...
-	TotalMetricsHistogram    uint64
-	TotalMetricsDistribution uint64
-	TotalMetricsSet          uint64
-	TotalMetricsTiming       uint64
-	TotalEvents              uint64
-	TotalServiceChecks       uint64
-	TotalDroppedOnReceive    uint64
+// statsdTelemetry contains metrics about the client
+type statsdTelemetry struct {
+	totalMetrics             uint32
```
truthbk, line 215: ditto about `uint32` vs `uint64`?

statsd/statsd.go
```diff
@@ -196,7 +196,7 @@ type Client struct {
 	// tags are global tags to be added to every statsd call
 	tags           []string
 	flushTime      time.Duration
-	metrics        *ClientMetrics
+	metrics        *statsdTelemetry
```
truthbk, line 199: nit: I think we'd have some more consistency if `metrics` were `telemetry` and `telemetry` were `telemetryCli` or something. Feel free to ignore.

statsd/telemetry.go
```diff
@@ -97,47 +192,85 @@ func (t *telemetryClient) sendTelemetry() {
...
+	return tlm
+}
+
 // flushTelemetry returns Telemetry metrics to be flushed. It's its own function to ease testing.
 func (t *telemetryClient) flush() []metric {
 	m := []metric{}
 
 	// same as Count but without global namespace
-	telemetryCount := func(name string, value int64, tags []string) {
+	telemetryGauge := func(name string, value int64, tags []string) {
```
truthbk, line 232: Since we're still sending an actual `count`, why the rename?

statsd/sender.go
truthbk: Ah OK, I think I see what's going on, we send the delta between flushes as a count so `uint32` should be enough to accomodate for that.

truthbk (commented):
All very minor comments, this looks pretty much good to go. My main concern was the change in type from `uint32` to `uint64`, but I can see how that should be enough between flush iterations.

statsd/sender.go
hush-hush: also, `count` are sent as float64, so a `uint64` would not fit and we need a type working with atomic. That why I fall back on uint32.

statsd/telemetry.go
hush-hush: left over from previous test using gauge

statsd/aggregator.go
```diff
@@ -124,22 +114,20 @@ func (a *aggregator) flush() {
...
 
-	am := &aggregatorMetrics{
-		nbContextGauge:        atomic.SwapInt32(&a.nbContextGauge, 0),
-		nbContextCount:        atomic.SwapInt32(&a.nbContextCount, 0),
-		nbContextSet:          atomic.SwapInt32(&a.nbContextSet, 0),
-		nbContextHistogram:    a.histograms.resetAndGetNbContext(),
-		nbContextDistribution: a.distributions.resetAndGetNbContext(),
-		nbContextTiming:       a.timings.resetAndGetNbContext(),
-	}
+	t.aggEnabled = true
```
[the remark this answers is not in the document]
hush-hush: I agree, not sure what we could do to improve it. WDYT ?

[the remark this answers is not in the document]
truthbk: I see how this plays well into some of the lock-less stuff you're doing, like the nil-pointer receiver checks.... Given that aggregation cannot be changed on the fly if I'm not mistaken, you could also set that up at initialization, but I guess this is fine otherwise.

truthbk (approved):

PR #225, opened 2021-09-03
Fix race condition in aggregation

hush-hush:
New metrics could overwrite each other when sampled for the first time.

### DIFF
--- a/statsd/aggregator.go
+++ b/statsd/aggregator.go
@@ -205,6 +205,13 @@ func (a *aggregator) count(name string, value int64, tags []string) error {
 	a.countsM.RUnlock()
 
 	a.countsM.Lock()
+	// Check if another goroutines hasn't created the value betwen the RUnlock and 'Lock'
+	if count, found := a.counts[context]; found {
+		count.sample(value)
+		a.countsM.Unlock()
+		return nil
+	}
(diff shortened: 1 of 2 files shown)


statsd/aggregator.go
```diff
@@ -223,6 +230,12 @@ func (a *aggregator) gauge(name string, value float64, tags []string) error {
 	gauge := newGaugeMetric(name, value, tags)
```
remeh, line 230: Silly question since I didn't dug much into datadog-go but, are you sure that you are winning anything from this `RUnlock` if you are re-locking the `gaugesM` right after that `newGaugeMetric` call? That question is true for all your changes since it's the same everywhere.

hush-hush: We need to `RUnlock` to be able to `Lock` right after.

remeh: My question is more about: is it worth unlocking then? Keeping the lock while running `newGaugeMetric` may not negatively impact performances since you actually try to get that lock right after again.

remeh: Discussed offline: the first RLock lets other routine sample the same metrics, so it can't be switched to Lock for performances reasons.

PR #226, opened 2021-09-03
Enable basic client side aggregation by default.

hush-hush:
Client side aggregation is no longer in BETA.


statsd/telemetry_test.go
```diff
@@ -107,33 +27,50 @@ func TestTelemetryCustomAddr(t *testing.T) {
 		WithNamespace("test_namespace"),
 	)
 
-	server := getTestServer(t, telAddr)
+	udpAddr, err := net.ResolveUDPAddr("udp", telAddr)
+	require.Nil(t, err, fmt.Sprintf("could not resolve udp '%s': %s", telAddr, err))
+	server, err := net.ListenUDP("udp", udpAddr)
+	require.Nil(t, err, fmt.Sprintf("Could not listen to UDP addr: %s", err))
```
matheusmosca, line 33: Nit:
```suggestion
	require.Nil(t, err, fmt.Sprintf("could not listen to UDP addr: %s", err))
```

PR #230, opened 2021-10-04
Fix go mod for v5 release

soh335:
Currently we cant use v5 via go mod. We need to define v5 at go.mod


hush-hush (approved):

hush-hush:
Thanks for the PR @soh335

PR #241, opened 2021-11-03
Fix missing import

anajavi:
uds_windows.go is missing "io" import resulting in build failure.

PR #242, opened 2021-11-16
Fix Windows compatibility for v5

programmer04:
Currently, version v5 of the library cannot be used on **Windows**, let's try
```
go test ./statsd
```
ends
```
# github.com/DataDog/datadog-go/v5/statsd [github.com/DataDog/datadog-go/v5/statsd.test]
statsd\statsd.go:264:25: too many arguments in call to newUDSWriter
        have (string, time.Duration)
        want (string)
statsd\uds_windows.go:8:33: undefined: io
FAIL    github.com/DataDog/datadog-go/v5/statsd [build failed]
FAIL
```
This PR fixes it, by
- adding missing `import io` (change has been already proposed by https://github.com/DataDog/datadog-go/pull/241, thanks @evanj)
- updating arguments expected by `newUDSWriter` for **Windows**
- increasing timeouts (I run it on flaky **Windows** VM with limited resources and original timeouts weren't sufficient, I've seen similar fix https://github.com/DataDog/datadog-go/pull/239)

In the future could you add CI testing on **Windows** too? It'll allow catching quickly such things before releases and save your time. Moreover, I recommend including macOS too, even if the library hasn't had this problem. I know that both systems are mostly Unix like and should be compatible, but some differences may exist (e.g. I stumbled in my project on the fact that Linux has syscall [gettid](https://man7.org/linux/man-pages/man2/gettid.2.html) and **macOS** doesn't). Errors may appear in the future so it's better to be extra cautious.

### DIFF
--- a/statsd/pipe_windows_test.go
+++ b/statsd/pipe_windows_test.go
@@ -87,6 +87,7 @@ func TestPipeWriterReconnect(t *testing.T) {
 
 	out := make(chan string)
 	go acceptOne(t, ln, out)
+
 	client, err := New(pipepath)
 	require.Nil(t, err)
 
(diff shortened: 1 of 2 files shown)


hush-hush (approved):

hush-hush:
Thanks for the PR !

PR #243, opened 2021-11-16
Fix windows compilation

Julio-Guerra:
Fix the Windows compilation error due to `newUDSWriter()` having a wrong type for the windows build tag.

Tested with:
```console
$ env GOOS=windows go build -v ./statsd/...
```


programmer04:
Duplicates #242

Issue #223, opened 2021-08-09
Support of `DD_TAGS` environment variables


afarbos:
Should this library support `DD_TAGS` like `dd-trace-go` do it [here](https://github.com/DataDog/dd-trace-go/blob/v1.32.0/ddtrace/tracer/option.go#L158) ?

This would avoid to use the [`WithTags`](https://pkg.go.dev/github.com/DataDog/datadog-go/statsd?utm_source=godoc#WithTags) and enable configuration from the environment.

SophieDeBenedetto:
👋 Hello Datadog friends! GitHub is interested in seeing this feature added to the datadog-go package, to give us parity with dogstatsd-ruby and to enable us to solve for some internal metric attribution problems we've been wrestling with. Would maintainers here be open to a contribution from us to deliver this functionality? I see that a related PR was commented on with some mild support, then closed in July of 2023 without comment.

Thanks for any insight you can provide!

PR #245, opened 2021-12-01
DD_TAGS env var support

pedro-stanaka:
## Summary

This PR fixes issue #223 and it is something that I am also interested in using. I hope this is the correct implementation for the problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,7 @@ env:
 ```
 
 * `DD_ENV`, `DD_SERVICE`, and `DD_VERSION` can be used by the statsd client to set `{env, service, version}` as global tags for all data emitted.
+* `DD_TAGS`: you can use this env var to send tags globally in all the emitted metrics. The value must be space separated (e.g. `DD_TAGS=custom_tag:foo other_tag:bar`).
 
 ### Unix Domain Sockets Client
 
--- a/statsd/end_to_end_udp_test.go
+++ b/statsd/end_to_end_udp_test.go
@@ -20,31 +20,40 @@ func TestPipelineWithGlobalTags(t *testing.T) {
 	ts.sendAllAndAssert(t, client)
 }
 
-func TestKownEnvTags(t *testing.T) {
+func TestKnownEnvTags(t *testing.T) {
 	entityIDEnvName := "DD_ENTITY_ID"
 	ddEnvName := "DD_ENV"
 	ddServiceName := "DD_SERVICE"
 	ddVersionName := "DD_VERSION"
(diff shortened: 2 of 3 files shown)


pedro-stanaka:
@hush-hush I know you had self-assigned issue #223 to you, so maybe you want to give me a review here? Thanks in advance!

hush-hush:
Hi @pedro-stanaka ,

Thanks for opening this PR. Adding support for `DD_TAGS` across all our DogStatsD clients is something we're looking into.

The main issue being that when `DD_TAGS` is set for the whole host and the agent/client are running locally it will be picked by both, so we have to be careful. This is especially important when using containers, VMs or k8s deployments. The last thing we want is for the clients or Agent to assign the wrong tags to a metric by default.

We're also looking into aligning the behavior between all clients around this, which is a bit all over the place right now.

In the meantime, clients that want to fetch tags from the env can easily do so using the library constructor. It's not the best but it's an easy workaround.

SophieDeBenedetto:
> Hi @pedro-stanaka ,
> 
> Thanks for opening this PR. Adding support for `DD_TAGS` across all our DogStatsD clients is something we're looking into.
> 
> The main issue being that when `DD_TAGS` is set for the whole host and the agent/client are running locally it will be picked by both, so we have to be careful. This is especially important when using containers, VMs or k8s deployments. The last thing we want is for the clients or Agent to assign the wrong tags to a metric by default.
> 
> We're also looking into aligning the behavior between all clients around this, which is a bit all over the place right now.
> 
> In the meantime, clients that want to fetch tags from the env can easily do so using the library constructor. It's not the best but it's an easy workaround.

I'm not sure I understand why there is a concern about telling the datadog-go SDK to use `DD_TAGS` may conflict with setting `DD_TAGS` at the agent level? In this approach, the datadog-go SDK looks for a `DD_TAGS` environment variable available to the container running the application code using the SDK. Any tags found there are applied to all metrics emitted by the SDK.

There is also an existing option to [configure a containerized datadog-agent running in Docker or similar to have `DD_TAGS` env var](https://docs.datadoghq.com/getting_started/tagging/assigning_tags/?tab=noncontainerizedenvironments) which would tell the agent to apply any such tags to all metrics it sends out to Datadog.

So, the way I understand this is:

* If you set `DD_TAGS` in the application environment, the datadog-go SDK will attach _those_ tags to all metrics it emits
* If you _also_ set `DD_TAGS` at the agent level, the agent will _also_ add _those_ tags to all metrics it emits

In other words, you get both sets of tags added globally.

Is this not desirable?

PR #172, opened 2020-10-26
[dogstatsd] Add support for including container id

xornivore:
- Adds `DD_ENTITY_TYPE` to enable collecting `container_id://..` entity ID automatically.
- Code for the cgroup parsing is essentially the same as in [dd-trace-go](https://github.com/DataDog/dd-trace-go/blob/afab3a8fcca25d2a37bf9de8870cd8295c840980/internal/container.go#L35).

PR #250, opened 2022-01-20
feat: Support container ID field (dogstatsd 1.2)

ahmed-mez:
**The PR does the following**
- Adds `c:<id>` field holding the container ID so the agent can attach container tags.

**Brief explanation of how the new container field is handled**
- If `DD_ENTITY_ID` is set, we ignore the container field
- If `DD_ENTITY_ID` is not set and `DD_ORIGIN_DETECTION_ENABLED` is not explicitly set to `false` or the `WithoutOriginDetection()` is not invoked, we try to get the app container ID by parsing `/proc/self/cgroup`. In case of success, the container field in will be set to <container ID>.

**Motivation**
- More granular tags (container tags instead of pod tags)
- Works for containers outside of Kubernetes (e.g ECS, Fargate)
- The goal of not reusing `dd.internal.entity_id` is to avoid parsing the value and guess the prefix on the agent side (avoid costly operations in the hot path)

**Notes**
- Users should explicitly enable container tagging **in the agent** otherwise the agent ignores the container field. So even if the client sends the field automatically, the tagging won't be done without explicitly enabling it in the agent.
- Cgroup parsing logic is essentially the same as in [dd-trace-go](https://github.com/DataDog/dd-trace-go/blob/v1/internal/container.go)
- This feature requires an agent change https://github.com/DataDog/datadog-agent/pull/10659
- Replaces https://github.com/DataDog/datadog-go/pull/172

### DIFF
--- a/statsd/container.go
+++ b/statsd/container.go
@@ -27,8 +27,8 @@ var (
 	// expContainerID matches contained IDs and sources. Source: https://github.com/Qard/container-info/blob/master/index.js
 	expContainerID = regexp.MustCompile(fmt.Sprintf(`(%s|%s|%s)(?:.scope)?$`, uuidSource, containerSource, taskSource))
 
-	userProvidedContainerID = ""
-	cgroupContainerID       = ""
+	// containerID holds the container ID.
+	containerID = ""
 )
 
 // parseContainerID finds the first container ID reading from r and returns it.
@@ -61,22 +61,22 @@ func readContainerID(fpath string) string {
 // It can either be auto-discovered with origin detection or provided by the user.
 // User-defined container ID is prioritized.
 func getContainerID() string {
-	if userProvidedContainerID != "" {
-		return userProvidedContainerID
-	}
-	return cgroupContainerID
+	return containerID
 }
 
-var readOnce sync.Once
+var initOnce sync.Once
 
-// setCgroupContainerID attempts to read the container ID from /proc/self/cgroup.
-func setCgroupContainerID() {
-	readOnce.Do(func() {
-		cgroupContainerID = readContainerID(cgroupPath)
+// initContainerID initializes the container ID.
+// It can either be provided by the user or read from cgroups.
+func initContainerID(userProvidedID string, cgroupFallback bool) {
+	initOnce.Do(func() {
+		if userProvidedID != "" {
+			containerID = userProvidedID
+			return
+		}
+
+		if cgroupFallback {
+			containerID = readContainerID(cgroupPath)
+		}
 	})
-}
-
-// setUserProvidedContainerID stores the container ID provided by the user.
-func setUserProvidedContainerID(id string) {
-	userProvidedContainerID = id
 }
--- a/statsd/container.go
+++ b/statsd/container.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"regexp"
+	"sync"
 )
 
 const (
@@ -25,6 +26,9 @@ var (
 
 	// expContainerID matches contained IDs and sources. Source: https://github.com/Qard/container-info/blob/master/index.js
 	expContainerID = regexp.MustCompile(fmt.Sprintf(`(%s|%s|%s)(?:.scope)?$`, uuidSource, containerSource, taskSource))
+
+	userProvidedContainerID = ""
+	cgroupContainerID       = ""
 )
 
 // parseContainerID finds the first container ID reading from r and returns it.
@@ -53,7 +57,26 @@ func readContainerID(fpath string) string {
 	return parseContainerID(f)
 }
 
-// getContainerID attempts to return the container ID from /proc/self/cgroup or empty on failure.
-var getContainerID func() string = func() string {
-	return readContainerID(cgroupPath)
+// getContainerID returns the container ID configured at the client creation
+// It can either be auto-discovered with origin detection or provided by the user.
+// User-defined container ID is prioritized.
+func getContainerID() string {
+	if userProvidedContainerID != "" {
+		return userProvidedContainerID
+	}
+	return cgroupContainerID
+}
+
+var readOnce sync.Once
+
+// setCgroupContainerID attempts to read the container ID from /proc/self/cgroup.
+func setCgroupContainerID() {
+	readOnce.Do(func() {
+		cgroupContainerID = readContainerID(cgroupPath)
+	})
+}
+
+// setUserProvidedContainerID stores the container ID provided by the user.
+func setUserProvidedContainerID(id string) {
+	userProvidedContainerID = id
 }
--- a/statsd/container.go
+++ b/statsd/container.go
@@ -1,0 +1,59 @@
+package statsd
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"regexp"
+)
+
+const (
+	// cgroupPath is the path to the cgroup file where we can find the container id if one exists.
+	cgroupPath = "/proc/self/cgroup"
+)
+
+const (
+	uuidSource      = "[0-9a-f]{8}[-_][0-9a-f]{4}[-_][0-9a-f]{4}[-_][0-9a-f]{4}[-_][0-9a-f]{12}"
+	containerSource = "[0-9a-f]{64}"
+	taskSource      = "[0-9a-f]{32}-\\d+"
+)
+
+var (
+	// expLine matches a line in the /proc/self/cgroup file. It has a submatch for the last element (path), which contains the container ID.
+	expLine = regexp.MustCompile(`^\d+:[^:]*:(.+)$`)
+
+	// expContainerID matches contained IDs and sources. Source: https://github.com/Qard/container-info/blob/master/index.js
+	expContainerID = regexp.MustCompile(fmt.Sprintf(`(%s|%s|%s)(?:.scope)?$`, uuidSource, containerSource, taskSource))
+)
+
+// parseContainerID finds the first container ID reading from r and returns it.
+func parseContainerID(r io.Reader) string {
+	scn := bufio.NewScanner(r)
+	for scn.Scan() {
+		path := expLine.FindStringSubmatch(scn.Text())
+		if len(path) != 2 {
+			// invalid entry, continue
+			continue
+		}
+		if parts := expContainerID.FindStringSubmatch(path[1]); len(parts) == 2 {
+			return parts[1]
+		}
+	}
+	return ""
+}
+
+// readContainerID attempts to return the container ID from the provided file path or empty on failure.
+func readContainerID(fpath string) string {
+	f, err := os.Open(fpath)
+	if err != nil {
+		return ""
+	}
+	defer f.Close()
+	return parseContainerID(f)
+}
+
+// getContainerID attempts to return the container ID from /proc/self/cgroup or empty on failure.
+var getContainerID func() string = func() string {
+	return readContainerID(cgroupPath)
+}
--- a/statsd/container_test.go
+++ b/statsd/container_test.go
@@ -1,0 +1,58 @@
+package statsd
+
+import (
+	"io"
+	"io/ioutil"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseContainerID(t *testing.T) {
+	for input, expectedResult := range map[string]string{
+		`other_line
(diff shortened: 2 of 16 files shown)


statsd/container_test.go
```diff
@@ -0,0 +1,65 @@
...
+		"1:name=systemd:/ecs/34dc0b5e626f2c5c4c5170e34b10e765-1234567890":                                                       "34dc0b5e626f2c5c4c5170e34b10e765-1234567890",
+		"1:name=systemd:/docker/34dc0b5e626f2c5c4c5170e34b10e7654ce36f0fcd532739f4445baabea03376.scope":                         "34dc0b5e626f2c5c4c5170e34b10e7654ce36f0fcd532739f4445baabea03376",
+		`1:name=systemd:/nope
+2:pids:/docker/34dc0b5e626f2c5c4c5170e34b10e7654ce36f0fcd532739f4445baabea03376
+3:cpu:/invalid`: "34dc0b5e626f2c5c4c5170e34b10e7654ce36f0fcd532739f4445baabea03376",
+	} {
+		id := parseContainerID(strings.NewReader(in))
+		if id != out {
+			t.Fatalf("%q -> %q: %q", in, out, id)
+		}
```
hush-hush, line 39:
```suggestion
                require.Equal(t, out, id)
```

statsd/container_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"io/ioutil"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseContainerID(t *testing.T) {
+	for in, out := range map[string]string{
```
hush-hush, line 14: Could we rename the variables:

```suggestion
	for input, expectedResult := range map[string]string{
```

statsd/end_to_end_udp_test.go
```diff
@@ -9,6 +9,12 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func dummyContainerID() string { return "" }
+
+func init() {
+	containerID = dummyContainerID
+}
+
```
hush-hush, line 17: This will impact every single test of the package, which means that by default we won't test the real code.

statsd/container.go
```diff
@@ -0,0 +1,59 @@
...
+func readContainerID(fpath string) string {
+	f, err := os.Open(fpath)
+	if err != nil {
+		return ""
+	}
+	defer f.Close()
+	return parseContainerID(f)
+}
+
+// ContainerID attempts to return the container ID from /proc/self/cgroup or empty on failure.
```
hush-hush, line 56:
```suggestion
// containerID attempts to return the container ID from /proc/self/cgroup or empty on failure.
```

statsd/container.go
```diff
@@ -0,0 +1,59 @@
...
+var containerID func() string = func() string {
```
hush-hush, line 57: Could we rename this `getContainerID` ?

statsd/end_to_end_udp_test.go
```diff
@@ -106,6 +112,70 @@ func TestKnownEnvTagsEmptyString(t *testing.T) {
...
+	)
+
+	sort.Strings(client.tags)
+	assert.Equal(t, expectedTags, client.tags)
+	ts.sendAllAndAssert(t, client)
+}
+
+func TestPrefixedOriginWithoutEntityID(t *testing.T) {
+	entityIDEnvName := "DD_ENTITY_ID"
+	os.Unsetenv(entityIDEnvName)
```
hush-hush, line 139:
```suggestion
	os.Unsetenv("DD_ENTITY_ID")
```

And same line 157

statsd/statsd.go
```diff
@@ -319,13 +340,29 @@ func newWithWriter(w io.WriteCloser, o *Options, writerName string) (*Client, er
...
+			}
 			c.tags = append(c.tags, fmt.Sprintf("%s:%s", mapping.tagName, value))
 		}
 	}
 
+	// If DD_ENTITY_ID is set, we don't send a prefixed origin ID
+	// as dd.internal.entity_id is prioritized over dd.internal.prefixed_origin
+	if !hasEntityID {
+		// Sending a prefixed origin can be disabled by configuring DD_UDP_ORIGIN_DISABLED=true
+		if isDisabled := strings.ToLower(os.Getenv(udpOriginDisabled)); isDisabled != "true" {
```
hush-hush, line 359: We could use https://pkg.go.dev/strconv#ParseBool here I think.

Also could we add an option for this (see the options.go file). And lastly, could we reverse the logic: instead of checking if it's disabled we check if it's enabled ?

Are we sure we want to enable this by default ? This would be a breaking change in behavior for customers (a new tag being sent that has an impact on the Agent behavior/performances).

statsd/end_to_end_udp_test.go
```diff
@@ -106,6 +112,70 @@ func TestKnownEnvTagsEmptyString(t *testing.T) {
...
+		"localhost:8765",
+		expectedTags,
+	)
+
+	sort.Strings(client.tags)
+	assert.Equal(t, expectedTags, client.tags)
+	ts.sendAllAndAssert(t, client)
+}
+
+func TestPrefixedOriginDisabled(t *testing.T) {
```
hush-hush, line 156: Coudl we add a check where both `DD_ENTITY_ID` and `DD_UDP_ORIGIN_DISABLED` are set ?

statsd/statsd.go
```diff
@@ -69,15 +69,36 @@ const (
...
+
+	// ddPrefixedOrigin specifies the tag name for the client-side prefixed entity ID injection.
+	// Currently the only supported format is dd.internal.prefixed_origin:container_id://<id>
+	// dd.internal.entity_id is prioritized over dd.internal.prefixed_origin for backward compatibility.
+	// If DD_ENTITY_ID is not set, the client tries to fill dd.internal.prefixed_origin automatically unless
+	// DD_UDP_ORIGIN_DISABLED is explicitly set to "true".
+	ddPrefixedOrigin = "dd.internal.prefixed_origin"
+
+	// udpOriginDisabled specifies the env var to disable injecting the dd.internal.prefixed_origin tag.
+	udpOriginDisabled = "DD_UDP_ORIGIN_DISABLED"
```
hush-hush, line 90: This logic could be use in the same way for UDS. Could we rename this to remove the `UDP` part ? Something like `DD_ORIGIN_DETECTION` ?

statsd/statsd.go
ahmed-mez: > We could use https://pkg.go.dev/strconv#ParseBool here I think.

> Also could we add an option for this (see the options.go file). And lastly, could we reverse the logic: instead of checking if it's disabled we check if it's enabled ?

Thanks! done!

> Are we sure we want to enable this by default ? This would be a breaking change in behavior for customers (a new tag being sent that has an impact on the Agent behavior/performances).

It's not enabled by default in the agent (see https://github.com/DataDog/datadog-agent/pull/10659). The agent doesn't even try to parse the field if the feature is not enabled. Having a single config in the agent will offer better UX.
cc @vickenty for visibility as this was your recommendation.

statsd/container.go
```diff
@@ -0,0 +1,59 @@
...
+		return ""
+	}
+	defer f.Close()
+	return parseContainerID(f)
+}
+
+// getContainerID attempts to return the container ID from /proc/self/cgroup or empty on failure.
+var getContainerID func() string = func() string {
+	return readContainerID(cgroupPath)
+}
```
hush-hush, line 59: Since the containerID is never going to change, could we save it in a var at the package level and initialized it only once with `sync.Once`.

Then `getContainerID` will just return that variable. This would avoid passing the container ID everywhere in the code. We could just call `getContainerID` in `appendContainerID` (in the format.go file). We could then control the value of that variable for easy testing.

This would simplify the code by a lot and avoid passing the same string every time.

WDYT ?

statsd/options.go
```diff
@@ -297,3 +300,15 @@ func WithTelemetryAddr(addr string) Option {
 		return nil
 	}
 }
+
+// WithoutOriginDetection disables the client origin detection.
+// When enabled, the client tries to discover its container ID and sends it to the Agent
+// to enrich the metrics with container tags.
```
hush-hush, line 306: We also need a `WithOriginDetection`, even if it's enabled by default. Some users have helpers that set some default settings and then each team/project can overwrite them. So this could be disabled by default and then enabled manually for a specific project. We would need tests for both option too.

Also, do we have a documentation we could link on how we discover the container ID ? If so, it would be nice to link it here. If not we should explain a bit more how we discover it.

Lastly, let's explain that it does not work on Windows and how `DD_ENTITY_ID` and `DD_ORIGIN_DETECTION_ENABLED` impact it.

hush-hush (changes requested):

statsd/container.go
ahmed-mez: > Since the containerID is never going to change, could we save it in a var at the package level and initialized it only once with sync.Once.

👍 

> Then getContainerID will just return that variable. This would avoid passing the container ID everywhere in the code. We could just call getContainerID in appendContainerID (in the format.go file). We could then control the value of that variable for easy testing.
> This would simplify the code by a lot and avoid passing the same string every time.
> WDYT ?

Sounds good, the reason why I implemented it that way was because I wanted to keep it consistent with the existing code and how metric properties are passed to the buffering logic.

ahmed-mez: Yes good point, so you're suggesting adding a `WithoutContainerID(id string)` option to the builder in addition to `WithoutOriginDetection`.

vickenty: Yes, except I think we should name them with a positive preposition: `WithContainerID` and `WithOriginDetection`.

ahmed-mez: I tried to apply all your suggestions in the latest commit, lmk how it looks :)

re `WithOriginDetection` we're going to provide both with and without, see https://github.com/DataDog/datadog-go/pull/250#discussion_r795581889

hush-hush: Yes we offer both `With` and `Without` in the client.

statsd/options.go
```diff
@@ -297,3 +301,43 @@ func WithTelemetryAddr(addr string) Option {
...
+//
+// More on this here: https://docs.datadoghq.com/developers/dogstatsd/?tab=kubernetes#origin-detection-over-udp
+func WithoutOriginDetection() Option {
+	return func(o *Options) error {
+		o.originDetection = false
+		return nil
+	}
+}
+
+// WithOriginDetection enables the client origin detection.
```
hush-hush, line 320: nit-pick: we should mention the version of the agent where this is supported (ie `7.35.0` I think ?)

hush-hush (changes requested):
I think it's almost good to go.

statsd/options.go
```diff
@@ -297,3 +301,45 @@ func WithTelemetryAddr(addr string) Option {
 		return nil
 	}
 }
+
+// WithoutOriginDetection disables the client origin detection.
+// This feature requires Datadog Agent 7.35.0+
```
hush-hush, line 306:
```suggestion
// WithoutOriginDetection disables the client origin detection.
// This feature requires Datadog Agent version >=6.35.0 && <7.0.0 or Agent's versions >=7.35.0.
```

I think we should move this to `WithOriginDetection` disabling it doesn't require any version of the agent.

statsd/options.go
```diff
@@ -297,3 +301,45 @@ func WithTelemetryAddr(addr string) Option {
...
+// More on this here: https://docs.datadoghq.com/developers/dogstatsd/?tab=kubernetes#origin-detection-over-udp
+func WithOriginDetection() Option {
+	return func(o *Options) error {
+		o.originDetection = true
+		return nil
+	}
+}
+
+// WithContainerID allows passing the container ID, this will be used by the Agent to enrich metrics with container tags.
+// This feature requires Datadog Agent 7.35.0+
```
hush-hush, line 337: same:
```suggestion
// This feature requires Datadog Agent version version >=6.35.0 && <7.0.0 or Agent's versions >=7.35.0.
```

hush-hush (changes requested):

statsd/options.go
ahmed-mez: indeed, done!

hush-hush (approved):
Thanks for the PR !

ahmed-mez:
Thank you for the review @hush-hush ! I'd like to get 👀  on the Agent PR before merging this one.

PTAL when you get a chance! https://github.com/DataDog/datadog-agent/pull/10659

vickenty (approved):

PR #252, opened 2022-02-08
Export internal error as `MessageTooLong`

### LINKED ISSUE #251: how to: handling "statsd buffer is full" error?
How do callers have to handle [statsd buffer is full](https://github.com/DataDog/datadog-go/blob/master/statsd/buffer.go#L11) error?

ATM a retry loop does the job, but I wonder if i'm missing anything since the error is private and presumably not supposed to bubble up all the way to the caller.

```go
// retry this error until the buffer is flushed
for i := 3; i > 0; i -= 1 {
	if err := d.Client.ServiceCheck(&sc); err == nil {
		break
	} else {
		// the error is private - comparing strings
		if err.Error() == "statsd buffer is full" {
			time.Sleep(1 * time.Second)
			continue
		}
		d.Logger.Error(err.Error())
		break
	}
}
```
Thank!

hush-hush:
This allows users to check the error type when sending message that can't fit in a buffer.

Resolve #251

### DIFF
--- a/statsd/buffer.go
+++ b/statsd/buffer.go
@@ -4,11 +4,15 @@ import (
 	"strconv"
 )
 
-type bufferFullError string
+// MessageTooLongError is an error returned when a sample, event or service check is too large once serialized. See
+// WithMaxBytesPerPayload option for more details.
+type MessageTooLongError struct{}
 
-func (e bufferFullError) Error() string { return string(e) }
+func (e MessageTooLongError) Error() string {
+	return "message too long. See 'WithMaxBytesPerPayload' documentation."
+}
 
-const errBufferFull = bufferFullError("statsd buffer is full")
+var errBufferFull = MessageTooLongError{}
 
 type partialWriteError string
 
--- a/statsd/options.go
+++ b/statsd/options.go
@@ -117,10 +117,13 @@ func WithMaxMessagesPerPayload(maxMessagesPerPayload int) Option {
 	}
 }
 
-// WithMaxBytesPerPayload sets the maximum number of bytes a single payload can contain.
+// WithMaxBytesPerPayload sets the maximum number of bytes a single payload can contain. Each sample, even and service
+// check must be lower than this value once serialized or an `MessageTooLongError` is returned.
 //
-// The deault value 0 which will set the option to the optimal size for the transport protocol used: 1432 for UDP and
-// named pipe and 8192 for UDS.
+// The default value 0 which will set the option to the optimal size for the transport protocol used: 1432 for UDP and
+// named pipe and 8192 for UDS. Those values offer the best performances.
+// Be careful when changing this option, see
+// https://docs.datadoghq.com/developers/dogstatsd/high_throughput/#ensure-proper-packet-sizes.
 func WithMaxBytesPerPayload(MaxBytesPerPayload int) Option {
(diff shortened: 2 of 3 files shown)


statsd/buffer.go
```diff
@@ -4,11 +4,15 @@ import (
...
+// WithMaxBytesPerPayload option for more details.
+type MessageTooLongError struct{}
 
-func (e bufferFullError) Error() string { return string(e) }
+func (e MessageTooLongError) Error() string {
+	return "message too long. See 'WithMaxBytesPerPayload' documentation."
+}
 
-const errBufferFull = bufferFullError("statsd buffer is full")
+var errBufferFull = &MessageTooLongError{}
```
ogaca-dd, line 15: Doesn't need to be a pointer.

statsd/options.go
```diff
@@ -117,10 +117,13 @@ func WithMaxMessagesPerPayload(maxMessagesPerPayload int) Option {
 	}
 }
 
-// WithMaxBytesPerPayload sets the maximum number of bytes a single payload can contain.
+// WithMaxBytesPerPayload sets the maximum number of bytes a single payload can contain. Each sample, even and service
+// check must be lower that this value once serialized or an `MessageTooLongError` will be returned.
```
ogaca-dd, line 121:
```suggestion
// check must be lower than this value once serialized or an `MessageTooLongError` is returned.
```

statsd/options.go
```diff
@@ -117,10 +117,13 @@ func WithMaxMessagesPerPayload(maxMessagesPerPayload int) Option {
...
 //
-// The deault value 0 which will set the option to the optimal size for the transport protocol used: 1432 for UDP and
-// named pipe and 8192 for UDS.
+// The default value 0 which will set the option to the optimal size for the transport protocol used: 1432 for UDP and
+// named pipe and 8192 for UDS. Those value offer the best performances.
```
ogaca-dd, line 124:
```suggestion
// named pipe and 8192 for UDS. Those values offer the best performances.
```

statsd/statsd_test.go
```diff
@@ -330,3 +330,15 @@ func Test_isOriginDetectionEnabled(t *testing.T) {
...
 	}
 }
+
+func TestMessageTooLongError(t *testing.T) {
+	client, err := New("localhost:8765", WithMaxBytesPerPayload(10), WithoutClientSideAggregation())
+	require.NoError(t, err)
+
+	err = client.Gauge("fake_name_", 21, nil, 1)
+	require.Error(t, err)
+	assert.Equal(t, "message too long. See 'WithMaxBytesPerPayload' documentation.", err.Error())
```
ogaca-dd, line 340: Isn't enough to check `err` type is `MessageTooLongError`?

statsd/buffer.go
hush-hush: Since we return this error often I wanted to avoid a copy each time.

ogaca-dd: `MessageTooLongError` is an empty struct so its size is `0`.

ogaca-dd (approved):

statsd/buffer.go
hush-hush: Indeed ! Forgot we move the error string :+1:

PR #253, opened 2022-02-11
Optimize getContext and getContextAndTags

martin-sucha:
It is not necessary to do multiple allocations and copying,
single pass is enough.

Screenshot from 2022-02-11 20-14-49
Screenshot from 2022-02-11 20-06-06

### DIFF
--- a/statsd/aggregator.go
+++ b/statsd/aggregator.go
@@ -172,12 +172,31 @@ func (a *aggregator) flushMetrics() []metric {
 }
 
 func getContext(name string, tags []string) string {
-	return name + ":" + strings.Join(tags, tagSeparatorSymbol)
+	c, _ := getContextAndTags(name, tags)
+	return c
 }
 
 func getContextAndTags(name string, tags []string) (string, string) {
-	stringTags := strings.Join(tags, tagSeparatorSymbol)
-	return name + ":" + stringTags, stringTags
+	n := len(name) + len(nameSeparatorSymbol) + len(tagSeparatorSymbol)*(len(tags)-1)
+	for _, s := range tags {
+		n += len(s)
+	}
+
+	var sb strings.Builder
+	sb.Grow(n)
+	sb.WriteString(name)
+	sb.WriteString(nameSeparatorSymbol)
+	if len(tags) > 0 {
+		sb.WriteString(tags[0])
+		for _, s := range tags[1:] {
+			sb.WriteString(tagSeparatorSymbol)
+			sb.WriteString(s)
+		}
+	}
+
+	s := sb.String()
+
+	return s, s[len(name)+len(nameSeparatorSymbol):]
 }
 
 func (a *aggregator) count(name string, value int64, tags []string) error {
--- a/statsd/aggregator_test.go
+++ b/statsd/aggregator_test.go
@@ -226,3 +226,42 @@ func TestAggregatorFlushConcurrency(t *testing.T) {
 
 	wg.Wait()
 }
+
+func TestGetContextAndTags(t *testing.T) {
+	tests := []struct {
+		testName    string
+		name        string
+		tags        []string
+		wantContext string
+		wantTags    string
+	}{
+		{
+			testName:    "no tags",
+			name:        "name",
+			tags:        nil,
+			wantContext: "name:",
+			wantTags:    "",
+		},
+		{
+			testName:    "one tag",
+			name:        "name",
+			tags:        []string{"tag1"},
+			wantContext: "name:tag1",
+			wantTags:    "tag1",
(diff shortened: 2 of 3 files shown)


statsd/aggregator.go
```diff
@@ -172,12 +172,31 @@ func (a *aggregator) flushMetrics() []metric {
...
 func getContext(name string, tags []string) string {
-	return name + ":" + strings.Join(tags, tagSeparatorSymbol)
+	c, _ := getContextAndTags(name, tags)
+	return c
 }
 
 func getContextAndTags(name string, tags []string) (string, string) {
-	stringTags := strings.Join(tags, tagSeparatorSymbol)
-	return name + ":" + stringTags, stringTags
+	n := len(name) + len(nameSeparatorSymbol) + len(tagSeparatorSymbol)*(len(tags)-1)
```
hush-hush, line 180: Doing a quick benchmark on `getContextAndTags` I get nice improvements except when we have 0 tags:

```
$> benchcmp before after
benchmark                           old ns/op     new ns/op     delta
BenchmarkGetContextAndTags0-6       64.9          118           +81.29%
BenchmarkGetContextAndTags1-6       89.0          74.0          -16.88%
BenchmarkGetContextAndTags5-6       297           157           -47.21%
BenchmarkGetContextAndTags10-6      428           259           -39.50%
BenchmarkGetContextAndTags50-6      1917          1186          -38.13%
BenchmarkGetContextAndTags100-6     3692          2358          -36.13%
```

If we catch this use case with by adding:
```
        if len(tags) == 0 {
		return name + ":", ""
	}
```

We get better performances everywhere:
```
$> benchcmp before after
benchmark                           old ns/op     new ns/op     delta
BenchmarkGetContextAndTags0-6       64.9          52.1          -19.72%
BenchmarkGetContextAndTags1-6       89.0          65.0          -26.93%
BenchmarkGetContextAndTags5-6       297           152           -48.99%
BenchmarkGetContextAndTags10-6      428           244           -43.01%
BenchmarkGetContextAndTags50-6      1917          1205          -37.14%
BenchmarkGetContextAndTags100-6     3692          2234          -39.49%
```

Could you add this edge case ?

hush-hush (changes requested):
Thanks a lot for this PR @martin-sucha !

I added a comment about a edge case. Beside this it looks ready to be merged.

statsd/aggregator.go
martin-sucha: Nice catch! Thanks for checking the benchmarks. I have added a commit to fix the edge case.

statsd/aggregator.go
```diff
@@ -172,12 +172,34 @@ func (a *aggregator) flushMetrics() []metric {
...
+	for _, s := range tags {
+		n += len(s)
+	}
+
+	var sb strings.Builder
+	sb.Grow(n)
+	sb.WriteString(name)
+	sb.WriteString(nameSeparatorSymbol)
+	if len(tags) > 0 {
```
hush-hush, line 192: We already catch this use case in the if at the top now.

martin-sucha: Oops, missed that. Thanks for noticing! Fixed.

hush-hush (approved):
Thanks for the PR !

I'll merge and will release it with the next version of the client (hopefully next week).

I'm curious of your use case and what bring you to optimizing this part of the client. Would you mind sharing how many points per second you're sending, what type of metrics ...

martin-sucha:
> I'm curious of your use case and what bring you to optimizing this part of the client.

We switched away from datadog-go several years ago to another statsd library because datadog-go did not have aggregation support at the time and was spending too much time sending packets. Now datadog-go has aggregation support, so I was checking if we can switch back to datadog-go as the other library does not support distribution metrics, which I'd like to use in some places. As part of that experiment, I profiled both versions.

As you can see in the image from the profiler in the original post, datadog-go was about 2.2% CPU time in the staging environment and getContext/getContextAndTags was majority of that. At the same time it was obvious from the flamegraph that the function can be optimized pretty easily. Now the profiler shows about 1.6% CPU for datadog-go.

I also tried Prometheus Go client (with counter vectors only) for comparison and that is around 1.4% CPU, so much closer now.

> Would you mind sharing how many points per second you're sending, what type of metrics ...

In the staging environment where I tested this about 42k metrics per second in one pod before aggregation (as shown by `datadog.dogstatsd.client.metrics_by_type` metric):

| --- | --- |
| Metric type | Rate |
| --- | --- |
| counter | 34k / second |
| timing | 7k / second |
| histogram | 800 / second |
| gauge | 30 / second |
| set | 0 |
| distribution | 0 |

Issue #140, opened 2020-03-25
How do I check whether Client has been closed?


jrefior:
It appears to be necessary to call Close() on statsd Clients to clean up cpu/memory resources.

How do I check whether a Client has been closed, so I can get a new client if needed?

kiyor-cai-exa:
I'm getting the same issue. I integrated to a custom logger and use a huge amount of events with a lot of different logger instances. The MEM resource is not able to release. But the same logic with Write to WebSocket and file (same time) are all able to close when doing GC. it did not happen half years ago I think was about version less than v2.3.0 .

jrefior:
I implemented a wrapper type with `Close()` and `IsClosed() bool` methods and that seems to be working well so far.

PR #254, opened 2022-02-23
feat: add IsClosed method statsd client

lucassscaravelli:
fix issues https://github.com/DataDog/datadog-go/issues/140

### DIFF
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -199,6 +199,9 @@ type ClientInterface interface {
 
 	// Flush forces a flush of all the queued dogstatsd payloads.
 	Flush() error
+
+	// IsClosed returns if the client connection is already closed
+	IsClosed() bool
 }
 
 // A Client is a handle for sending messages to dogstatsd.  It is safe to
@@ -223,6 +226,7 @@ type Client struct {
 	aggExtended     *aggregator
 	options         []Option
 	addrOption      string
+	isClosed        bool
 }
 
 // statsdTelemetry contains telemetry metrics about the client
@@ -470,6 +474,11 @@ func (c *Client) Flush() error {
 	return nil
 }
 
+// IsClosed returns if the cliente connection is closed.
+func (c *Client) IsClosed() bool {
+	return c.isClosed
+}
+
 func (c *Client) flushTelemetryMetrics(t *Telemetry) {
 	t.TotalMetricsGauge = atomic.LoadUint64(&c.telemetry.totalMetricsGauge)
 	t.TotalMetricsCount = atomic.LoadUint64(&c.telemetry.totalMetricsCount)
@@ -648,6 +657,10 @@ func (c *Client) Close() error {
 		return ErrNoClient
 	}
 
+	if c.isClosed {
+		return nil
+	}
+
 	// Acquire closer lock to ensure only one thread can close the stop channel
 	c.closerLock.Lock()
 	defer c.closerLock.Unlock()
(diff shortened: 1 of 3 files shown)


statsd/statsd.go
```diff
@@ -648,6 +657,10 @@ func (c *Client) Close() error {
 		return ErrNoClient
 	}
 
+	if c.isClosed {
+		return nil
+	}
+
```
hush-hush, line 663: Could we move this after the lock below

statsd/statsd_test.go
```diff
@@ -165,6 +165,23 @@ func TestCloseRace(t *testing.T) {
...
 }
 
+func TestCloseWithClientAlreadyClosed(t *testing.T) {
+	c, err := New("localhost:8125")
+	c.isClosed = true
+	assert.NoError(t, err)
+	c.Close()
+	assert.True(t, c.IsClosed())
+	c.sender.close()
+}
```
hush-hush, line 175: I think testing a double `Close` would be better:
```suggestion
func TestCloseWithClientAlreadyClosed(t *testing.T) {
	c, err := New("localhost:8125")
	assert.NoError(t, err)
	assert.False(t, c.IsClosed())
	
	assert.NoError(t, c.Close())
	assert.True(t, c.IsClosed())
	
	assert.NoError(t, c.Close())
        assert.True(t, c.IsClosed())
}
```

statsd/statsd_test.go
```diff
@@ -165,6 +165,23 @@ func TestCloseRace(t *testing.T) {
...
+
+func TestIsClosed(t *testing.T) {
+	c, err := New("localhost:8125")
+	assert.NoError(t, err)
+	assert.False(t, c.IsClosed())
+	c.Close()
```
hush-hush, line 181: Could we test that no error is returned too ?

statsd/statsd.go
```diff
@@ -678,7 +691,13 @@ func (c *Client) Close() error {
...
 
 	c.Flush()
-	return c.sender.close()
+
+	if err := c.sender.close(); err != nil {
+		return err
+	}
+
+	c.isClosed = true
+	return nil
```
hush-hush, line 700: Since we closed the workers and aggregators we can consider that the client is closed anyway event if the sender returns an error (as it's not longer capable or receiving points).

```suggestion
        c.isClosed = true
        return c.sender.close()
```

statsd/statsd.go
```diff
@@ -470,6 +474,11 @@ func (c *Client) Flush() error {
 	return nil
 }
 
+// IsClosed returns if the cliente connection is closed.
```
hush-hush, line 477:
```suggestion
// IsClosed returns if the client has been closed.
```

statsd/statsd.go
```diff
@@ -199,6 +199,9 @@ type ClientInterface interface {
 
 	// Flush forces a flush of all the queued dogstatsd payloads.
 	Flush() error
+
+	// IsClosed returns if the client connection is already closed
```
hush-hush, line 203:
```suggestion
	// IsClosed returns if the client has been closed.
```

hush-hush:
Thanks a lot for the PR @lucassscaravelli !

I added a few comments, but overall I think we will be able to merge.

hush-hush (changes requested):

lucassscaravelli:
@hush-hush thanks for review! Fixed 👍

statsd/statsd.go
```diff
@@ -470,6 +474,11 @@ func (c *Client) Flush() error {
 	return nil
 }
 
+// IsClosed returns if the client has been closed.
+func (c *Client) IsClosed() bool {
+	return c.isClosed
```
hush-hush, line 479: I missed this during the first review, but we need to  lock `c.closerLock` here too.

hush-hush (changes requested):
One last change

statsd/statsd.go
lucassscaravelli: Done 👍

hush-hush (approved):

PR #255, opened 2022-03-01
Adding mock interface to ease user testing

hush-hush:


statsd/noop.go
```diff
@@ -86,6 +86,11 @@ func (n *NoOpClient) IsClosed() bool {
 	return false
 }
 
+// GetTelemetry does nothing and returns an emtpy Telemetry
```
olivielpeau, line 89:
```suggestion
// GetTelemetry does nothing and returns an empty Telemetry
```

.github/workflows/generate.yaml
```diff
@@ -0,0 +1,24 @@
+
+# We use github actions to test the code on windows and linux amd64. Circleci is used for linux arm64.
+
```
olivielpeau, line 3: This comment should probably be removed

```suggestion
```

.github/workflows/generate.yaml
```diff
@@ -0,0 +1,24 @@
...
+      fail-fast: false
+    runs-on: ${{ matrix.runs-on }}
+    steps:
+      - name: Setup go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - run: go install github.com/golang/mock/mockgen@v1.6.0
```
olivielpeau, line 22: I don't think you need to do this since `go generate` is a module-aware command? https://go.dev/ref/mod#build-commands

(and I think removing it would remove the need to update the version pinned here when the version pinned in `go.mod` is updated)

olivielpeau (approved):
Just a few suggestions, but as long as CI passes no need for further review from me :) Thanks for doing this!

.github/workflows/generate.yaml
olivielpeau: I was confused here: we still need to install the package. But we should avoid pinning it here as well

As a conclusion to our a convo on slack, let's replace this with:

```suggestion
      - run: go install github.com/golang/mock/mockgen
```
so that we install the package at the version defined in `go.mod`.

olivielpeau (approved):

PR #260, opened 2022-06-15
Ensure that the atomic uint64 will always be aligned

jdgordon:
I'm not sure if this is a bug in the go compiler, or if this struct is being embedded somewhere which is breaking alignment rules, but on ARM atomic access to the nbContext field causes a panic due to unaligned access. Switching to a pointer ensures it will always be correctly aligned.

I have been running this branch on my device for the last 2 weeks without any other issues.

### DIFF
--- a/statsd/buffered_metric_context.go
+++ b/statsd/buffered_metric_context.go
@@ -11,7 +11,7 @@ import (
 // and Timing. Since those 3 metric types behave the same way and are sampled
 // with the same type they're represented by the same class.
 type bufferedMetricContexts struct {
-	nbContext uint64
+	nbContext *uint64
 	mutex     sync.RWMutex
 	values    bufferedMetricMap
 	newMetric func(string, float64, string) *bufferedMetric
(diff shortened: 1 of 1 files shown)


sgnn7:
Hey @jdgordon,
We're [aware of this issue](https://github.com/DataDog/datadog-agent/blob/main/docs/dev/atomics.md) and have started fixing it in [some spots](https://github.com/DataDog/datadog-agent/pull/11985) but we haven't gotten to this code yet. Let me see if I can find the right person to take a look at this and decide if we want to move to `go.uber.com/atomic` for this code as well.

djmitche:
Thanks!  We're standardizing on go.uber.org/atomic for the datadog-agent itself, so it would make sense to do the same here.  It should be an easy change to this PR.

DylanRJohnston-FZ:
We've hit this issue in production and will have to use this fork for now.

katiehockman:
Thanks for your patience with this. @knusbaum has a [PR out now](https://github.com/DataDog/dd-trace-go/pull/1443) that should fix the issue for you. There are just a few more changes that need to happen on that PR before it can be merged.

knusbaum:
The fix for this issue was merged and released in 1.42.0. I'm going to close this now, but thank you for the contribution.

PR #261, opened 2022-06-21
Carlosroman/go 1 18

carlosroman:


.github/workflows/datadog-go.yaml
```diff
@@ -1,16 +1,35 @@
...
 
 jobs:
   native:
     strategy:
       matrix:
-        go-version: [ 1.17, 1.16, 1.15, 1.14, 1.13]
-        runs-on: [ ubuntu-latest, windows-latest, macos-latest ]
-      fail-fast: false
+        go-version:
+          - 1.13
```
obalunenko, line 21: As far as 1.12 version is [mentioned here](https://github.com/DataDog/datadog-go/pull/261/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R256) it will be nice to add it to the matrix?

carlosroman: Yeah, the reason this hasn't been merged is because the unit tests won't run for `1.12`. I'll hopefully have some time to look at this and fix the matrix + update our support plan.

PR #262, opened 2022-07-28
statsd: support sending metrics with timestamp.

remeh:
Adds two methods `GaugeWithTimestamp` and `CountWithTimestamp` in the `ClientInterface` to support sending gauge and count with timestamp, for users aggregating metrics on their side and who can provide a timestamp for a given metric.

Metrics sent using `GaugeWithTimestamp` and `CountWithTimestamp` to the DogStatsD server will be directly sent to the intake with no aggregation done server-side.

If the client is created using `WithClientSideAggregation()`, no aggregation will be performed on metrics submitted with timestamp, they will be written in the serialization buffer as-is.

### DIFF
--- a/statsd/aggregator.go
+++ b/statsd/aggregator.go
@@ -14,9 +14,6 @@ type (
 	bufferedMetricMap map[string]*bufferedMetric
 )
 
-// noTimestamp is used as a value for metric without a given timestamp.
-const noTimestamp = int64(0)
-
 type aggregator struct {
 	nbContextGauge uint64
 	nbContextCount uint64
--- a/statsd/aggregator.go
+++ b/statsd/aggregator.go
@@ -221,15 +221,7 @@ func (a *aggregator) count(name string, value int64, tags []string) error {
 		return nil
 	}
 
-	a.counts[context] = newCountMetric(name, value, tags, noTimestamp)
-	a.countsM.Unlock()
-	return nil
-}
-
-func (a *aggregator) countWithTimestamp(name string, value int64, tags []string, timestamp int64) error {
-	context := getContext(name, tags)
-	a.countsM.Lock()
-	a.counts[context] = newCountMetric(name, value, tags, timestamp)
+	a.counts[context] = newCountMetric(name, value, tags)
 	a.countsM.Unlock()
 	return nil
 }
@@ -244,7 +236,7 @@ func (a *aggregator) gauge(name string, value float64, tags []string) error {
 	}
 	a.gaugesM.RUnlock()
 
-	gauge := newGaugeMetric(name, value, tags, noTimestamp)
+	gauge := newGaugeMetric(name, value, tags)
 
 	a.gaugesM.Lock()
 	// Check if another goroutines hasn't created the value betwen the 'RUnlock' and 'Lock'
@@ -254,14 +246,6 @@ func (a *aggregator) gauge(name string, value float64, tags []string) error {
 		return nil
 	}
 	a.gauges[context] = gauge
-	a.gaugesM.Unlock()
-	return nil
-}
-
-func (a *aggregator) gaugeWithTimestamp(name string, value float64, tags []string, timestamp int64) error {
-	context := getContext(name, tags)
-	a.gaugesM.Lock()
-	a.gauges[context] = newGaugeMetric(name, value, tags, timestamp)
 	a.gaugesM.Unlock()
 	return nil
 }
--- a/statsd/aggregator.go
+++ b/statsd/aggregator.go
@@ -14,6 +14,9 @@ type (
 	bufferedMetricMap map[string]*bufferedMetric
 )
 
+// noTimestamp is used as a value for metric without a given timestamp.
+const noTimestamp = int64(0)
+
 type aggregator struct {
 	nbContextGauge uint64
 	nbContextCount uint64
@@ -218,7 +221,15 @@ func (a *aggregator) count(name string, value int64, tags []string) error {
 		return nil
 	}
 
-	a.counts[context] = newCountMetric(name, value, tags)
+	a.counts[context] = newCountMetric(name, value, tags, noTimestamp)
+	a.countsM.Unlock()
+	return nil
+}
+
+func (a *aggregator) countWithTimestamp(name string, value int64, tags []string, timestamp int64) error {
+	context := getContext(name, tags)
+	a.countsM.Lock()
+	a.counts[context] = newCountMetric(name, value, tags, timestamp)
 	a.countsM.Unlock()
 	return nil
 }
@@ -233,7 +244,7 @@ func (a *aggregator) gauge(name string, value float64, tags []string) error {
 	}
 	a.gaugesM.RUnlock()
 
-	gauge := newGaugeMetric(name, value, tags)
+	gauge := newGaugeMetric(name, value, tags, noTimestamp)
 
 	a.gaugesM.Lock()
 	// Check if another goroutines hasn't created the value betwen the 'RUnlock' and 'Lock'
@@ -243,6 +254,14 @@ func (a *aggregator) gauge(name string, value float64, tags []string) error {
 		return nil
 	}
 	a.gauges[context] = gauge
+	a.gaugesM.Unlock()
+	return nil
+}
+
+func (a *aggregator) gaugeWithTimestamp(name string, value float64, tags []string, timestamp int64) error {
+	context := getContext(name, tags)
+	a.gaugesM.Lock()
+	a.gauges[context] = newGaugeMetric(name, value, tags, timestamp)
 	a.gaugesM.Unlock()
 	return nil
 }
--- a/statsd/buffer.go
+++ b/statsd/buffer.go
@@ -39,22 +39,28 @@ func newStatsdBuffer(maxSize, maxElements int) *statsdBuffer {
 	}
 }
 
-func (b *statsdBuffer) writeGauge(namespace string, globalTags []string, name string, value float64, tags []string, rate float64) error {
+func (b *statsdBuffer) writeGauge(namespace string, globalTags []string, name string, value float64, tags []string, rate float64, timestamp int64) error {
 	if b.elementCount >= b.maxElements {
 		return errBufferFull
 	}
 	originalBuffer := b.buffer
 	b.buffer = appendGauge(b.buffer, namespace, globalTags, name, value, tags, rate)
+	if timestamp != noTimestamp {
+		b.buffer = appendTimestamp(b.buffer, timestamp)
+	}
 	b.writeSeparator()
 	return b.validateNewElement(originalBuffer)
 }
 
-func (b *statsdBuffer) writeCount(namespace string, globalTags []string, name string, value int64, tags []string, rate float64) error {
+func (b *statsdBuffer) writeCount(namespace string, globalTags []string, name string, value int64, tags []string, rate float64, timestamp int64) error {
 	if b.elementCount >= b.maxElements {
 		return errBufferFull
 	}
 	originalBuffer := b.buffer
 	b.buffer = appendCount(b.buffer, namespace, globalTags, name, value, tags, rate)
+	if timestamp != noTimestamp {
+		b.buffer = appendTimestamp(b.buffer, timestamp)
+	}
 	b.writeSeparator()
 	return b.validateNewElement(originalBuffer)
 }
--- a/statsd/buffer.go
+++ b/statsd/buffer.go
@@ -45,9 +45,7 @@ func (b *statsdBuffer) writeGauge(namespace string, globalTags []string, name st
 	}
 	originalBuffer := b.buffer
 	b.buffer = appendGauge(b.buffer, namespace, globalTags, name, value, tags, rate)
-	if timestamp != noTimestamp {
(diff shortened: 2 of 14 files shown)


statsd/statsd.go
```diff
@@ -159,9 +160,19 @@ type ClientInterface interface {
 	// Gauge measures the value of a metric at a particular time.
 	Gauge(name string, value float64, tags []string, rate float64) error
 
+	// Gauge measures the value of a metric at a given time.
+	// Even with client side aggregation enabled, there is no aggregation done on a metric
+	// sent using GaugeWithTimestamp: it is written as is in the serialization buffer.
+	GaugeWithTimestamp(name string, value float64, tags []string, rate float64, timestamp time.Time) error
+
```
hush-hush, line 167: Why only `gauge` and `count` ? `rate`/`set` doesn't make send but we might want to send `timing`, `histogram` and `distribution` with timestamp too, no ?

statsd/statsd.go
```diff
@@ -159,9 +160,19 @@ type ClientInterface interface {
 	// Gauge measures the value of a metric at a particular time.
 	Gauge(name string, value float64, tags []string, rate float64) error
 
+	// Gauge measures the value of a metric at a given time.
+	// Even with client side aggregation enabled, there is no aggregation done on a metric
+	// sent using GaugeWithTimestamp: it is written as is in the serialization buffer.
```
hush-hush, line 165: I this this should be more high level doc. Users reading the API doc for the client might not understand/know all the options for the client.

WDYT:

```suggestion
	// GaugeWithTimestamp measures the value of a metric at a given time.
	/// The value will bypass any aggregation on the client side and agent side. The Agent will simply forward this
	/// metrics to the backend. This is useful when sending points in the past.
```

statsd/statsd.go
```diff
@@ -159,9 +160,19 @@ type ClientInterface interface {
...
+	GaugeWithTimestamp(name string, value float64, tags []string, rate float64, timestamp time.Time) error
+
 	// Count tracks how many times something happened per second.
 	Count(name string, value int64, tags []string, rate float64) error
 
+	// CountWithTimestamp tracks how many times something happened at the given second.
+	// Even with client side aggregation enabled, there is no aggregation done on a metric
+	// sent using CountWithTimestamp: it is written as is in the serialization buffer.
```
hush-hush, line 173: Same here

statsd/statsd.go
```diff
@@ -552,6 +563,17 @@ func (c *Client) Gauge(name string, value float64, tags []string, rate float64)
 	return c.send(metric{metricType: gauge, name: name, fvalue: value, tags: tags, rate: rate, globalTags: c.tags, namespace: c.namespace})
 }
 
+// Gauge measures the value of a metric at a given time.
```
hush-hush, line 566:
```suggestion
// GaugeWithTimestamp measures the value of a metric at a given time.
```

statsd/statsd.go
```diff
@@ -552,6 +563,17 @@ func (c *Client) Gauge(name string, value float64, tags []string, rate float64)
...
+// Even with client side aggregation enabled, there is no aggregation done on a metric
+// sent using GaugeWithTimestamp: it is written as is in the serialization buffer.
+func (c *Client) GaugeWithTimestamp(name string, value float64, tags []string, rate float64, timestamp time.Time) error {
```
hush-hush, line 569: Should we take a `time.Time` or a `int` directly here ?

statsd/statsd.go
```diff
@@ -564,6 +586,17 @@ func (c *Client) Count(name string, value int64, tags []string, rate float64) er
 	return c.send(metric{metricType: count, name: name, ivalue: value, tags: tags, rate: rate, globalTags: c.tags, namespace: c.namespace})
 }
 
+// Count tracks how many times something happened at the given second.
```
hush-hush, line 589:
```suggestion
// CountWithTimestamp tracks how many times something happened at the given second.
```

statsd/statsd.go
```diff
@@ -552,6 +563,17 @@ func (c *Client) Gauge(name string, value float64, tags []string, rate float64)
...
 }
 
+// Gauge measures the value of a metric at a given time.
+// Even with client side aggregation enabled, there is no aggregation done on a metric
+// sent using GaugeWithTimestamp: it is written as is in the serialization buffer.
+func (c *Client) GaugeWithTimestamp(name string, value float64, tags []string, rate float64, timestamp time.Time) error {
+	if c == nil {
+		return ErrNoClient
+	}
+	atomic.AddUint64(&c.telemetry.totalMetricsGauge, 1)
```
hush-hush, line 573: Do we want a dedicated telemetry for metrics without timestamp ? Do we want to know what % of traffic is with timestamp ?

hush-hush (changes requested):

statsd/statsd.go
hush-hush, line 569: We should return an error if timestamp is 0 (same for count).

hush-hush (changes requested):

statsd/aggregator.go
```diff
@@ -14,6 +14,9 @@ type (
 	bufferedMetricMap map[string]*bufferedMetric
 )
 
+// noTimestamp is used as a value for metric without a given timestamp.
+const noTimestamp = int64(0)
+
```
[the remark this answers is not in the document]
remeh: Good catch, moving it out into statsd.go

statsd/buffer.go
```diff
@@ -39,22 +39,28 @@ func newStatsdBuffer(maxSize, maxElements int) *statsdBuffer {
...
 }
 
-func (b *statsdBuffer) writeGauge(namespace string, globalTags []string, name string, value float64, tags []string, rate float64) error {
+func (b *statsdBuffer) writeGauge(namespace string, globalTags []string, name string, value float64, tags []string, rate float64, timestamp int64) error {
 	if b.elementCount >= b.maxElements {
 		return errBufferFull
 	}
 	originalBuffer := b.buffer
 	b.buffer = appendGauge(b.buffer, namespace, globalTags, name, value, tags, rate)
+	if timestamp != noTimestamp {
```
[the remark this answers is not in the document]
remeh: I'm kind of disagreeing with that because to me, it's when writing a gauge and a count (so, in `writeGauge` and `writeCount`) that we don't want to write a zero timestamp. While having `appendTimestamp` writing a `0` value is correct by meaning and should be possible. 
Also, your solution means we have this:
```
	b.buffer = appendGauge(b.buffer, namespace, globalTags, name, value, tags, rate)
	b.buffer = appendTimestamp(b.buffer, timestamp)
```
while no timestamp will be written if `timestamp == 0`, which makes this invisible to the reader.

But honestly, I've hesitated when implementing this and no strong opinion so I'll follow your idea.

statsd/statsd.go
remeh: For now, timed metrics ingestion will only be supported for Gauge and Count (DogStatsD server side and backend side).

statsd/statsd.go
remeh: Using your suggestion, thanks.

remeh: I just remove the " The Agent will simply forward this metrics to the Datadog." which is finally redundant with your previous sentence; if you don't mind.

statsd/statsd.go
remeh: I think we should go with `time.Time` just because it a public-facing API and `time.Time` is here for that, showing you _have_ to provide something which is representing time, not just an int. That's subtle but I think that's the way to go. WDYT?

statsd/statsd.go
remeh: For now I don't think we have a reason adding a new metrics (sent by the client to the intake by every clients...) just for this.
I think this is an open question that I have to solve: I open a ticket for this but I think the best thing to do is to use the same telemetry for now (it's not wrong neither to count it in the "total metrics gauge" and "total metrics count").

statsd/buffer.go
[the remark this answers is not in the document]
hush-hush: First I think this would avoid duplicating the logic of `if timestamp != noTimestamp {` in multiple places as well as making it future proof for other devs.

But from a design POV I think the `appendTimestamp` should encapsulate the logic around timestamp. `writeGauge` or `writeCount` should not have to know every implementation details of the protocol. all they know is the order of the building blocks (metric name then type then tags, ...).

statsd/statsd.go
hush-hush: I think we should explain that somewhere, maybe the README or the Go doc string.

statsd/statsd.go
hush-hush: Sounds good

statsd/statsd.go
hush-hush: I think it's an error since go has default value. This means an non-initialized timestamp. 
Sending a `time.Time` at 0 will produce a `Unix` timestamp of `-62135596800`.

remeh: A gotcha... Also linked to your other question about `time.Time` vs `uint64`...
You are right I'll add a test `if timestamp.Unix() < 0`, WDYT?

hush-hush (approved):

rafaelsierra (approved):

ogaca-dd (commented):
LGTM (Nothing to add compare to existing comments).

statsd/statsd.go
```diff
@@ -564,6 +607,24 @@ func (c *Client) Count(name string, value int64, tags []string, rate float64) er
 	return c.send(metric{metricType: count, name: name, ivalue: value, tags: tags, rate: rate, globalTags: c.tags, namespace: c.namespace})
 }
 
+// CountWithTimestamp tracks how many times something happened at the given second.
+// The value will bypass any aggregation on the client side and agent side.
+// This is useful when sending points in the past.
```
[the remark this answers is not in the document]
remeh: I think adding the supported granularity isn't such an addition as it is written:
`CountWithTimestamp tracks how many times something happened at the given second.`
But I'll revisit this when we start documenting the actual feature in public docs and such.

statsd/statsd.go
```diff
@@ -152,16 +156,37 @@ func (e noClientErr) Error() string {
...
 // purpose of being able to provide a no-op client or even mocking. This can aid
 // downstream users' with their testing.
 type ClientInterface interface {
 	// Gauge measures the value of a metric at a particular time.
 	Gauge(name string, value float64, tags []string, rate float64) error
 
+	// GaugeWithTimestamp measures the value of a metric at a given time.
+	// The value will bypass any aggregation on the client side and agent side.
+	// This is useful when sending points in the past.
+	// Minimum Datadog Agent version: 7.39.0
```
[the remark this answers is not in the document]
remeh: Correct!

statsd/statsd.go
```diff
@@ -126,6 +126,9 @@ const (
 	writerWindowsPipe string = "pipe"
 )
 
+// noTimestamp is used as a value for metric without a given timestamp.
+const noTimestamp = int64(0)
```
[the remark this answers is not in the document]
remeh: Good catch.
I think it may be simpler to change the `appendTimestamp` implementation to this test instead:
```
	if timestamp >= noTimestamp {
```
WDYT?

statsd/statsd.go
remeh: Yeah.. We have to draw a limit somewhere in the clients and since the intake has a "minimum timestamp" limit and we already have other places limiting to the unix epoch in the Agent, I think having `int(0)` as the minimal possible value is speaking to everyone in a client. Let me know if you think I'm wrong, really open to discussion with these topics.

> I assume this doesn't apply to us?

It doesn't indeed, a future PR will update the public documentation, with things more related about the "features" than the technical chunks in the client.

statsd/statsd.go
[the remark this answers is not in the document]
remeh: Switched to documented guidelines: `timestamp` must be > 0 to be appended to a message or to be considered valid by `GaugeWithTimestamp` and `CountWithTimestamp`.

statsd/statsd.go
```diff
@@ -552,6 +577,24 @@ func (c *Client) Gauge(name string, value float64, tags []string, rate float64)
...
+// The value will bypass any aggregation on the client side and agent side.
+// This is useful when sending points in the past.
+//
+// Minimum Datadog Agent version: 7.40.0
+func (c *Client) GaugeWithTimestamp(name string, value float64, tags []string, rate float64, timestamp time.Time) error {
+	if c == nil {
+		return ErrNoClient
+	}
+
+	if timestamp.Unix() < 0 {
```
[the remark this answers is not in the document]
remeh: I first introduced a `minValidTimestamp` const but it was only used here so I've decided to re-use `noTimestamp` instead, WDYT?

statsd/statsd.go
remeh: > I'm trying to understand why 0 (jan 1970) is the limit here.

The idea is to provide a hard-coded limit instead of no limit, even if it has no actual feature meaning here.
I added a mention in the comment to redirect to the feature documentation.
```
Please report to the Datadog documentation for the maximum age of a metric.
```

scottopell (approved):

PR #269, opened 2022-11-24
Bug fix: Call ticker.Stop() when aggregator is stopped

byronwolfman:
## TL;DR
This is a bugfix PR that prevents the aggregator's ticker from being leaked when the statsd client is closed.

## Longer version

For reasons that aren't super interesting, we have an app that periodically allocates and closes statsd clients, e.g.

```go
d, err := statsd.New("")
// Do some stuff
err = d.Close()
```

The app does this frequently enough that we discovered a leaked ticker. The DataDog continuous profiler helped zero in on the problem by hinting that the app was spending an increasing number of CPU cycles on `runtime.CheckTimers` and `runtime.netpoll`/`runtime.epollwait`. For a one-liner change, the effect is pretty outsized! You can spot the difference:

| Screen Shot 2022-11-24 at 1 31 20 PM |
|-|

| Screen Shot 2022-11-24 at 1 57 04 PM |
|-|

| Screen Shot 2022-11-24 at 1 33 29 PM |
|-|

| Screen Shot 2022-11-24 at 1 33 41 PM |
|-|

### DIFF
--- a/statsd/aggregator.go
+++ b/statsd/aggregator.go
@@ -66,6 +66,7 @@ func (a *aggregator) start(flushInterval time.Duration) {
 			case <-ticker.C:
 				a.flush()
 			case <-a.closed:
+				ticker.Stop()
 				return
 			}
 		}


scottopell (approved):
Thanks for fixing this and writing up a detailed PR, this change looks great.

PR #273, opened 2023-02-21
Adding support for DD_DOGSTATSD_URL

carlosroman:
To keep parity with `ddtrace` (the Datadog Python APM Client), I've added support for `DD_DOGSTATSD_URL` that behaves as described in `ddtrace` docs [here](https://ddtrace.readthedocs.io/en/stable/configuration.html#DD_DOGSTATSD_URL).

> The URL to use to connect the Datadog agent for Dogstatsd metrics. The url can start with `udp://` to connect using UDP or with `unix://` to use a Unix Domain Socket.
>
> Example for UDP url: `DD_DOGSTATSD_URL=udp://localhost:8125`
>
> Example for UDS: `DD_DOGSTATSD_URL=unix:///var/run/datadog/dsd.socket`
>

If you use `DD_DOGSTATSD_URL=udp://localhost` it will add the correct port (i.e. change it to `DD_DOGSTATSD_URL=udp://localhost:8125`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,8 +84,16 @@ Find a list of all the available options for your DogStatsD Client in the [Datad
 
 ### Supported environment variables
 
-* If the `addr` parameter is empty, the client uses the `DD_AGENT_HOST` environment variables to build a target address.
-  Example: `DD_AGENT_HOST=127.0.0.1:8125` for UDP, `DD_AGENT_HOST=unix:///path/to/socket` for UDS and `DD_AGENT_HOST=\\.\pipe\my_windows_pipe` for Windows named pipe.
+* If the `addr` parameter is empty, the client will:
+  * First use the `DD_DOGSTATSD_URL` environment variables to build a target address. This must be a URL that start with either `udp://` (to connect using UDP) or with `unix://` (to use a Unix Domain Socket).
+    Example for UDP url: `DD_DOGSTATSD_URL=udp://localhost:8125`
+    Example for UDS: `DD_DOGSTATSD_URL=unix:///var/run/datadog/dsd.socket`
+    Example for Windows named pipe`DD_AGENT_HOST=\\.\pipe\my_windows_pipe`
+  * Fallback to the `DD_AGENT_HOST` environment variables to build a target address.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -84,8 +84,13 @@ Find a list of all the available options for your DogStatsD Client in the [Datad
...
 
-* If the `addr` parameter is empty, the client uses the `DD_AGENT_HOST` environment variables to build a target address.
-  Example: `DD_AGENT_HOST=127.0.0.1:8125` for UDP, `DD_AGENT_HOST=unix:///path/to/socket` for UDS and `DD_AGENT_HOST=\\.\pipe\my_windows_pipe` for Windows named pipe.
+* If the `addr` parameter is empty, the client will:
+  * First use the `DD_DOGSTATSD_URL` environment variables to build a target address. The url can start with `udp://` to connect using UDP or with `unix://` to use a Unix Domain Socket.
+    Example for UDP url: `DD_DOGSTATSD_URL=udp://localhost:8125`
+    Example for UDS: `DD_DOGSTATSD_URL=unix:///var/run/datadog/dsd.socket`
+    Example for Windows named pipe`DD_AGENT_HOST=\\.\pipe\my_windows_pipe`
+  * Fallback to the `DD_AGENT_HOST` environment variables to build a target address.
+    Example: `DD_AGENT_HOST=127.0.0.1:8125` for UDP, `DD_AGENT_HOST=unix:///path/to/socket` for UDS and `DD_AGENT_HOST=\\.\pipe\my_windows_pipe` for Windows named pipe.
```
scottopell, line 93: Is the omission of `DD_AGENT_PORT` intentional here? Is it semi-deprecated?

carlosroman: It was never documented in the `READMe.md`. I'll add it with the refactor to remove the `goto`.

scottopell (approved):
Looks good to me! Should be included in our release notes for the next datadog-go version.

If an application was previously setting `DD_DOGSTATSD_URL` to some incorrect value and `DD_AGENT_HOST` to their desired value, then this upgrade would "break" them since `DD_DOGSTATSD_URL` takes precedence over `DD_AGENT_HOST`. Seems pretty unlikely, so I'm not worried about it being a breaking change.

PR #279, opened 2023-05-23
Use multiple UDP sockets for sending statsd packets.

tjmerritt:
Use multiple UDP sockets for sending statsd packets.

Currently the datadog-go statsd client uses a single UDP socket to send packets to the statsd server.  On nodes with a larger number of cores (64+) this creates a bottleneck for sending packets.  The Veneur statsd proxy can receive packets on multiple sockets, but since datadog-go only sends packets via a single socket, veneur will only receive packets on a single socket.  The single socket becomes a bottleneck within the Linux kernel, and the maximum packet rate is determined by rate that a single core can send packets.  This PR provides datadog-go the capability to send packets over multiple sockets and increase the maximum packet rate that the datadog-go client can emit

### DIFF
--- a/statsd/options.go
+++ b/statsd/options.go
@@ -24,6 +24,7 @@ var (
 	defaultAggregation              = true
 	defaultExtendedAggregation      = false
 	defaultOriginDetection          = true
+	defaultUDPSocketCount           = 1
 )
 
 // Options contains the configuration options for a client.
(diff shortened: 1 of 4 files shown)


carlosroman:
Going to close as we actually recommend using [DogStatsD over Unix Domain Socket](https://docs.datadoghq.com/extend/dogstatsd/unix_socket?tab=host) for sending large volumes of metrics as it is more efficient.

PR #280, opened 2023-05-26
refresh udp ip at some cadence

blakearnold:
Forked https://github.com/DataDog/datadog-go/pull/264, but put the refresh in a separate go routine for performance.

The current UDP connection logic will resolve an address only when the writer is created. This is undesirable in cases where the address may change at a later point.


natenho:
@AliDatadog @wdhif  Is there any chance this PR will be included in a release?

abennett:
Just ran into this issue today with a dns change that occurred against a remote instance of the datadog-agent.

PR #284, opened 2023-10-11
Give more error handling options the the applications

iksaif:
From the point of view of the applications metrics (and buffers) are currently silently dropped. It makes it hard to:
- Make issues observable to the application and user (in logs for example)
- Allow the application to adapt its traffic to what the agent can handle

This commit adds:
- A way to return an error when metrics are dropped early
- A way to define an error handler (like https://github.com/DataDog/java-dogstatsd-client/blob/master/src/main/java/com/timgroup/statsd/StatsDClientErrorHandler.java#L9)

By default both of these are disabled (for performance reasons) but would offer better observability in drops, their reason and how to act (change settings, etc..)

This is going to be increasingly necessary as multi-tenant environment will require back-pressure and drops will need to be more actionable.

### DIFF
--- a/statsd/buffer.go
+++ b/statsd/buffer.go
@@ -23,7 +23,7 @@ const errPartialWrite = partialWriteError("value partially written")
 const metricOverhead = 512
 
 // statsdBuffer is a buffer containing statsd messages
-// this struct methods are NOT safe for concurent use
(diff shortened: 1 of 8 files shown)


remeh (approved):
Elegant implementation, LGTM.

PR #286, opened 2023-10-13
Add support for AF_UNIX+SOCK_STREAM

iksaif:
This is an experimental transport that will allow to have one connection.

### DIFF
--- a/example/simple_example.go
+++ b/example/simple_example.go
@@ -2,13 +2,19 @@ package main
 
 import (
 	"log"
+	"os"
 
 	"github.com/DataDog/datadog-go/v5/statsd"
 )
 
 func main() {
-	client, err := statsd.New("127.0.0.1:8125",
+	addr := os.Getenv("DOGSTATSD_ADDR")
+	if addr == "" {
+		addr = "127.0.0.1:8125"
+	}
+	client, err := statsd.New(addr,
 		statsd.WithTags([]string{"env:prod", "service:myservice"}),
+		statsd.WithErrorHandler(statsd.LoggingErrorHandler),
 	)
 	if err != nil {
 		log.Fatal(err)
--- a/example/simple_example.go
+++ b/example/simple_example.go
@@ -2,19 +2,13 @@ package main
 
 import (
 	"log"
-	"os"
 
 	"github.com/DataDog/datadog-go/v5/statsd"
 )
 
 func main() {
-	addr := os.Getenv("DOGSTATSD_ADDR")
-	if addr == "" {
-		addr = "127.0.0.1:8125"
-	}
-	client, err := statsd.New(addr,
+	client, err := statsd.New("127.0.0.1:8125",
 		statsd.WithTags([]string{"env:prod", "service:myservice"}),
-		statsd.WithErrorHandler(statsd.LoggingErrorHandler),
 	)
 	if err != nil {
 		log.Fatal(err)
--- a/statsd/uds.go
+++ b/statsd/uds.go
@@ -33,9 +33,6 @@ func newUDSWriter(addr string, writeTimeout time.Duration, transport string) (*u
 
 // retryOnWriteErr returns true if we should retry writing after a write error
 func (w *udsWriter) retryOnWriteErr(err error) bool {
-	if err == nil {
-		return true
-	}
 	// Never retry when using unixgram (to preserve the historical behavior)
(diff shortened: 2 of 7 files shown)


statsd/uds.go
```diff
@@ -20,33 +25,90 @@ type udsWriter struct {
...
 	// Defer connection to first Write
-	writer := &udsWriter{addr: udsAddr, conn: nil, writeTimeout: writeTimeout}
+	writer := &udsWriter{addr: addr, transport: transport, conn: nil, writeTimeout: writeTimeout}
 	return writer, nil
 }
 
+// retryOnWriteErr returns true if we should retry writing after a write error
+func (w *udsWriter) retryOnWriteErr(err error) bool {
+	if err == nil {
+		return true
```
remeh, line 37: I think this is incorrect.

statsd/uds.go
```diff
@@ -20,33 +25,90 @@ type udsWriter struct {
...
+	}
+	return false
+}
+
+// writeFull writes the whole data to the UDS connection
+func (w *udsWriter) writeFull(data []byte, stopIfNoneWritten bool) (int, error) {
+	written := 0
+	for written < len(data) {
+		n, e := w.conn.Write(data)
```
remeh, line 62: Shouldn't you write `data[written:]`?

statsd/uds.go
```diff
@@ -20,33 +25,90 @@ type udsWriter struct {
...
+			return written, e
+		}
+
+		// If there's an error, check if it is retryable
+		if e != nil && !w.retryOnWriteErr(e) {
+			return written, e
+		}
+
+		// When using "unix" we need to be able to finish to write partially written packets once we have started.
+		if w.transport == "unix" {
```
remeh, line 76: Confidence check:
```
	w.transport = newConn.RemoteAddr().Network()
```
This always returns `unix`, never `unixstream` or similar? Otherwise this current line should be adapted.

statsd/uds.go
```diff
@@ -20,33 +25,90 @@ type udsWriter struct {
...
 	conn.SetWriteDeadline(time.Now().Add(w.writeTimeout))
-	n, e := conn.Write(data)
 
-	if err, isNetworkErr := e.(net.Error); err != nil && (!isNetworkErr || !err.Temporary()) {
-		// Statsd server disconnected, retry connecting at next packet
+	// When using streams, we append the length of the packet to the data
+	if w.transport == "unix" {
+		bs := []byte{0, 0, 0, 0}
+		binary.LittleEndian.PutUint32(bs, uint32(len(data)))
+		_, err = w.writeFull(bs, true)
```
remeh, line 100: There is no `if err != nil` error management here, should we do something when we can't write our header?

example/simple_example.go
```diff
@@ -2,13 +2,19 @@ package main
...
 import (
 	"log"
+	"os"
 
 	"github.com/DataDog/datadog-go/v5/statsd"
 )
 
 func main() {
-	client, err := statsd.New("127.0.0.1:8125",
+	addr := os.Getenv("DOGSTATSD_ADDR")
```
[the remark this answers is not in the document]
iksaif: ok will remove, I originally did this change to be able to try another address

statsd/uds.go
iksaif: removed this branch

statsd/uds.go
iksaif: 😭

statsd/uds.go
iksaif:
```
// Network returns the address's network name, "unix", "unixgram" or
// "unixpacket".
func (a *UnixAddr) Network() string {
	return a.Net
}
```

I think we're good ?

statsd/uds.go
iksaif:
```
		_, err = w.writeFull(bs, true)
		if err == nil {
			n, err = w.writeFull(data, false)
		}
```

There is because if we fail to write the header we skip the payload and move to error management a bit later

statsd/uds.go
remeh: I think it was ok to have it (I mean, what you tried to do with it `if err == nil { return false }`

remeh (approved):
LGTM! Something tricky is that we'll have to have a mention in the CHANGELOG that `unix://smthing` has a slight change in behavior, while it is still very experimental...

PR #289, opened 2023-11-07
buffered_metrics: reduce a bit lock contention

iksaif:
062e18cef60a416fb19f9203534c936198e79595 introduced a performance regression by trying to fix the computation of the actual sampling rate.

This commit takes some shortcuts to reduce lock contention while trying to keep the sampling rate correct.

### DIFF
--- a/statsd/buffered_metric_context.go
+++ b/statsd/buffered_metric_context.go
@@ -57,7 +57,8 @@ func (bc *bufferedMetricContexts) flush(metrics []metric) []metric {
 func (bc *bufferedMetricContexts) sample(name string, value float64, tags []string, rate float64) error {
 	keepingSample := shouldSample(rate, bc.random, &bc.randomLock)
 
-	// If we don't keep the sample, return early, if we do store *first* observed sampling rate in the metric.
+	// If we don't keep the sample, return early. If we do keep the sample
+	// we end up storing the *first* observed sampling rate in the metric.
 	// This is the *wrong* behavior but it's the one we had before and the alternative would increase lock contention too
 	// much with the current code.
 	// TODO: change this behavior in the future, probably by introducing thread-local storage and lockless stuctures.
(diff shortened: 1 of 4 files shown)


vickenty (dismissed):

statsd/buffered_metric_context.go
```diff
@@ -57,6 +57,15 @@ func (bc *bufferedMetricContexts) flush(metrics []metric) []metric {
 func (bc *bufferedMetricContexts) sample(name string, value float64, tags []string, rate float64) error {
 	keepingSample := shouldSample(rate, bc.random, &bc.randomLock)
 
+	// If we don't keep the sample, return early, if we do store *first* observed sampling rate in the metric.
```
scottopell, line 60: I'm having a hard time understanding this comment, here's a suggestion
```suggestion
	// If we don't keep the sample, return early. If we do keep the sample
	// we end up storing the *first* observed sampling rate in the metric.
```

statsd/options.go
```diff
@@ -312,6 +312,8 @@ func WithoutClientSideAggregation() Option {
 
 // WithExtendedClientSideAggregation enables client side aggregation for all types. This feature is only compatible with
 // Agent's version >=6.25.0 && <7.0.0 or Agent's versions >=7.25.0.
+// When enabled the use of `rate` with distribution is discouraged and `WithMaxSamplesPerContext()` should be used.
```
scottopell, line 315:
```suggestion
// When enabled, the use of `rate` with distribution is discouraged and `WithMaxSamplesPerContext()` should be used.
```

statsd/metrics.go
```diff
@@ -184,18 +189,30 @@ func (s *bufferedMetric) skipSample() {
...
 
 func (s *bufferedMetric) flushUnsafe() metric {
+	totalSample := atomic.LoadInt64(&s.totalSamples)
+	var rate float64
+
+	// If the user had a specified rate send it because we don't know better.
+	// This code should be removed once we can also remove the early return at the top of
+	// `bufferedMetricContexts.sample`
+	if s.specifiedRate != 1.0 {
+		rate = s.specifiedRate
```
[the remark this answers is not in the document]
iksaif: I'll update the comment, originally I planned to try to do something smarter, but I think it would require more thinking .. so reverting to the previous behavior is a better first step

vickenty (approved):

PR #290, opened 2023-11-27
statsd: set client_transport properly

iksaif:
This propagates the transport from the sender (and writer) to the telemetry client directly.

This introduces a new `Transport` interface with `getTransportName()` and teaches the telemetry client to read this and refresh its tags when necessary.

The result will be that:
- 'unix://` will use `client_transport:uds` unless it connects to a UDS stream socket (and then switch to `client_transport:uds-stream`)
- 'unixstream://` will always directly use `client_transport:uds-stream`
- No changes for the other writers

### DIFF
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -139,12 +139,11 @@ const (
 )
 
 const (
-	writerNameUDP         string = "udp"
-	writerNameUDS         string = "uds"
-	writerNameUDSDatagram string = "uds-datagram"
-	writerNameUDSStream   string = "uds-stream"
-	writerWindowsPipe     string = "pipe"
-	writerNameCustom      string = "custom"
+	writerNameUDP       string = "udp"
+	writerNameUDS       string = "uds"
+	writerNameUDSStream string = "uds-stream"
+	writerWindowsPipe   string = "pipe"
+	writerNameCustom    string = "custom"
 )
 
 // noTimestamp is used as a value for metric without a given timestamp.
@@ -381,10 +380,10 @@ func createWriter(addr string, writeTimeout time.Duration) (Transport, string, e
 		return w, writerNameUDS, err
 	case strings.HasPrefix(addr, UnixAddressDatagramPrefix):
 		w, err := newUDSWriter(addr[len(UnixAddressDatagramPrefix):], writeTimeout, "unixgram")
-		return w, writerNameUDSDatagram, err
+		return w, writerNameUDS, err
 	case strings.HasPrefix(addr, UnixAddressStreamPrefix):
 		w, err := newUDSWriter(addr[len(UnixAddressStreamPrefix):], writeTimeout, "unix")
-		return w, writerNameUDSStream, err
+		return w, writerNameUDS, err
 	default:
 		w, err := newUDPWriter(addr, writeTimeout)
 		return w, writerNameUDP, err
@@ -467,7 +466,7 @@ func newWithWriter(w Transport, o *Options, writerName string) (*Client, error) 
 		initContainerID(o.containerID, isOriginDetectionEnabled(o, hasEntityID))
 	}
 
-	isUDS := writerName == writerNameUDS || writerName == writerNameUDSDatagram || writerName == writerNameUDSStream
+	isUDS := writerName == writerNameUDS
 
 	if o.maxBytesPerPayload == 0 {
 		if isUDS {
--- a/statsd/telemetry.go
+++ b/statsd/telemetry.go
@@ -126,13 +126,6 @@ type telemetryClient struct {
 	lastSample        Telemetry // The previous sample of telemetry sent
 }
 
-func clientTransportTag(transport string) string {
-	if transport == writerNameUDS || transport == writerNameUDSDatagram {
-		return writerNameUDS // For backward compatibility.
-	}
-	return transport
-}
-
 func newTelemetryClient(c *Client, aggregationEnabled bool) *telemetryClient {
 	t := &telemetryClient{
 		c:          c,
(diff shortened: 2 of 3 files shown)


statsd/statsd.go
```diff
@@ -378,10 +380,10 @@ func createWriter(addr string, writeTimeout time.Duration) (io.WriteCloser, stri
 		return w, writerNameUDS, err
 	case strings.HasPrefix(addr, UnixAddressDatagramPrefix):
 		w, err := newUDSWriter(addr[len(UnixAddressDatagramPrefix):], writeTimeout, "unixgram")
-		return w, writerNameUDS, err
+		return w, writerNameUDSDatagram, err
```
vickenty, line 383: Why use a distinct name here, isn't the behavior the same as with unix socket with no scheme prefix?

statsd/statsd.go
```diff
@@ -456,22 +458,24 @@ func newWithWriter(w io.WriteCloser, o *Options, writerName string) (*Client, er
 		initContainerID(o.containerID, isOriginDetectionEnabled(o, hasEntityID))
 	}
 
+	isUDS := writerName == writerNameUDS || writerName == writerNameUDSDatagram || writerName == writerNameUdsStream
+
 	if o.maxBytesPerPayload == 0 {
-		if writerName == writerNameUDS {
+		if isUDS {
```
vickenty, line 464: Commit or PR message should mention this change to payload size as well.

iksaif: 🤔 this doesn't change anything no ? before `writerNameUDS` was used for `unix://` / `unixgram://` / `unixstream://`, now because we separate the writer name based on the protocol schema, we intoduce the `isUDS` variable. But nothing else changes

statsd/statsd.go
iksaif: It is, but this allows the `client_transport` tag to be correctly set by in telemetry client

statsd/statsd.go
iksaif: Updated the descrption to make this more obvious

vickenty:
> When unix:// is set, since we guess the name the type of the socket based on what is on the filesystem, we will use use client_transport:uds. Otherwise we will use client_trasnport:uds-stream or client_transport:uds-datagram

1. Would it be possible to have the telemetry tag reflect actual socket type and not just configuration?
2. Would it make sense to keep telemetry tags compatible with other clients and also backwards compatible? `uds` currently means "unix datagram sockets", can we keep it this way?

iksaif:
It's not really possible to do (1) because the tags are defined before the writer guesses the socket type.

The current implementation seems to be backward compatible with other clients and previous behavior in the sense that nothing will change for existing configurations (but more explicit configuration will allow us to have more explicit tagging).

In the long run I think it's a good incentive to have people move to `unixstream://` or `unixgram://` instead of just `unix://` to make it clear what is the expected socket type

statsd/uds.go
```diff
@@ -31,6 +31,18 @@ func newUDSWriter(addr string, writeTimeout time.Duration, transport string) (*u
...
+// GetTransportName returns the transport used by the writer
+func (w *udsWriter) GetTransportName() string {
+	if w.transport == "unixgram" {
+		return writerNameUDS
+	} else if w.transport == "unix" {
+		return writerNameUDSStream
+	} else {
+		// For backward compatibility.
+		return writerNameUDS
+	}
```
vickenty, line 43: I think this comment may be misleading, the only way this branch is taken if we'd have `unixpacket` transport, for which there is nothing to be backward compatible with.
```suggestion
	if w.transport == "unix" {
		return writerNameUDSStream
	} else {
		return writerNameUDS
	}
```

statsd/telemetry.go
```diff
@@ -113,39 +113,45 @@ type Telemetry struct {
...
+	worker            *worker
+	lastSample        Telemetry // The previous sample of telemetry sent
 }
 
-func newTelemetryClient(c *Client, transport string, aggregationEnabled bool) *telemetryClient {
+func clientTransportTag(transport string) string {
+	if transport == writerNameUDS || transport == writerNameUDSDatagram {
+		return writerNameUDS // For backward compatibility.
+	}
+	return transport
```
vickenty, line 133: This function seems redundant now, since we return correct transport tag directly from the client?

statsd/uds.go
```diff
@@ -31,6 +31,18 @@ func newUDSWriter(addr string, writeTimeout time.Duration, transport string) (*u
 	return writer, nil
 }
 
+// GetTransportName returns the transport used by the writer
+func (w *udsWriter) GetTransportName() string {
+	if w.transport == "unixgram" {
```
vickenty, line 36: I think `w.transport` access needs a lock to avoid a data race.

iksaif: done, and removed a few string comparisons

statsd/telemetry.go
iksaif: removed

statsd/uds.go
iksaif: applied

statsd/statsd.go
```diff
@@ -378,10 +381,10 @@ func createWriter(addr string, writeTimeout time.Duration) (io.WriteCloser, stri
 		return w, writerNameUDS, err
 	case strings.HasPrefix(addr, UnixAddressDatagramPrefix):
 		w, err := newUDSWriter(addr[len(UnixAddressDatagramPrefix):], writeTimeout, "unixgram")
-		return w, writerNameUDS, err
+		return w, writerNameUDSDatagram, err
 	case strings.HasPrefix(addr, UnixAddressStreamPrefix):
 		w, err := newUDSWriter(addr[len(UnixAddressStreamPrefix):], writeTimeout, "unix")
-		return w, writerNameUDS, err
+		return w, writerNameUDSStream, err
```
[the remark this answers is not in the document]
iksaif: Ok, simplified

vickenty (approved):

PR #296, opened 2023-12-14
Add an API to send multiple samples at once

iksaif:
DistributionSamples is similar to Distribution, but it lets the client deals with the sampling, rate is passed to the agent and not used for further sampling. WithMaxBufferedMetricsPerContext is ignored when using this method.

Because this is a shift compared to how other methods are behaving, this method is provided only in ClientDirect which provides direct access to some low level dogstatsd protocol features like [DataDog/datadog-go#extended-aggregation](https://github.com/DataDog/datadog-go#extended-aggregation)

This is recommended in high performance cases were the overhead of the statsd library might be significant and the sampling is already done by the client.

`NewDirect` and `NewDirectWithWriter` can be used to instantiate a new direct client, mocks and noop clients are available.

replaces https://github.com/DataDog/datadog-go/pull/292


statsd/statsd.go
```diff
@@ -215,6 +215,9 @@ type ClientInterface interface {
 	Histogram(name string, value float64, tags []string, rate float64) error
 
 	// Distribution tracks the statistical distribution of a set of values across your infrastructure.
+	//
+	// It is recommended to use `WithMaxBufferedMetricsPerContext` to avoid dropping metrics, `rate` can
```
olivielpeau, line 219:
```suggestion
	// It is recommended to use `WithMaxBufferedMetricsPerContext` to avoid dropping metrics at high throughput, `rate` can
```

vickenty (dismissed):

remeh (approved):

PR #297, opened 2023-12-15
aggregator: don't allocate memory if we don't have tags

iksaif:
Before:
`BenchmarkGetContextNoTags-10    	39929401	        29.70 ns/op	      16 B/op	       1 allocs/op`

After:
`BenchmarkGetContextNoTags-10    	419886585	         2.902 ns/op	       0 B/op	       0 allocs/op`

This might look not much, but this makes a lot of difference in term of allocation when a counter is called repeatedly inside a loop.

This also fixes the other benchmarks that I wanted to run to validate I didn't cause regression but where not able to run properly

We can see it in this profile (referencing line 182 explicitely):
Screenshot 2023-12-15 at 11 52 49

Referencing https://github.com/DataDog/datadog-go/blob/master/statsd/aggregator.go#L182 as the main cause of allocation

From a simple call to Count:
`statsd.Count("rollout.done", 1, nil, 1.0)`

### DIFF
--- a/statsd/aggregator.go
+++ b/statsd/aggregator.go
@@ -172,14 +172,22 @@ func (a *aggregator) flushMetrics() []metric {
 	return metrics
 }
 
+// getContext returns the context for a metric name and tags.
+//
+// The context is the metric name and tags separated by a separator symbol.
+// It is not intended to be used as a metric name but as a unique key to aggregate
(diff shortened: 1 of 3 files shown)


remeh (approved):
LGTM. Good catch.

PR #299, opened 2024-01-02
uds: implement a connect timeout option

iksaif:
This is similar to https://github.com/DataDog/java-dogstatsd-client/pull/228

The idea is to protect the client from starved connections (due to listener bugs) and reinitialise the connection as needed.

### DIFF
--- a/statsd/uds.go
+++ b/statsd/uds.go
@@ -21,13 +21,15 @@ type udsWriter struct {
 	conn net.Conn
 	// write timeout
 	writeTimeout time.Duration
-	sync.RWMutex // used to lock conn / writer can replace it
+	// connect timeout
+	connectTimeout time.Duration
+	sync.RWMutex   // used to lock conn / writer can replace it
 }
 
 // newUDSWriter returns a pointer to a new udsWriter given a socket file path as addr.
-func newUDSWriter(addr string, writeTimeout time.Duration, transport string) (*udsWriter, error) {
+func newUDSWriter(addr string, writeTimeout time.Duration, connectTimeout time.Duration, transport string) (*udsWriter, error) {
 	// Defer connection to first Write
-	writer := &udsWriter{addr: addr, transport: transport, conn: nil, writeTimeout: writeTimeout}
+	writer := &udsWriter{addr: addr, transport: transport, conn: nil, writeTimeout: writeTimeout, connectTimeout: connectTimeout}
 	return writer, nil
 }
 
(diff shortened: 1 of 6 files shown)


iksaif:
The failing test has been flaky for a while :/

statsd/uds.go
```diff
@@ -43,56 +45,23 @@ func (w *udsWriter) GetTransportName() string {
...
-func (w *udsWriter) shouldCloseConnection(err error) bool {
 	if err, isNetworkErr := err.(net.Error); err != nil && (!isNetworkErr || !err.Timeout()) {
 		// Statsd server disconnected, retry connecting at next packet
 		return true
 	}
 	return false
 }
 
-// writeFull writes the whole data to the UDS connection
-func (w *udsWriter) writeFull(data []byte, stopIfNoneWritten bool, stream bool) (int, error) {
```
vickenty, line 68: This doesn't seem to be related to connection timeout. Can this change be split in a separate PR?

iksaif: Like in the Java PR the `connectionTimeout` is used instead of the `writeTimeout` to reset the connection if a partial write timeouts.

Do you still think we should add this in a separate PR?

statsd/uds.go
```diff
@@ -107,15 +76,26 @@ func (w *udsWriter) Write(data []byte) (int, error) {
 	if stream {
 		bs := []byte{0, 0, 0, 0}
 		binary.LittleEndian.PutUint32(bs, uint32(len(data)))
-		_, err = w.writeFull(bs, true, true)
+		_, err = w.conn.Write(bs)
```
vickenty, line 79: Would it make sense to check number of bytes written? AFAIK Go documentation makes no statement that it the write will be complete if the error is nil.

statsd/uds.go
```diff
@@ -182,5 +162,6 @@ func (w *udsWriter) ensureConnection() (net.Conn, error) {
 func (w *udsWriter) unsetConnection() {
 	w.Lock()
 	defer w.Unlock()
+	_ = w.conn.Close()
```
vickenty, line 165: Would it make sense to log the error?

statsd/uds.go
iksaif: https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/io/io.go;l=99

```
// Write writes len(p) bytes from p to the underlying data stream.
// It returns the number of bytes written from p (0 <= n <= len(p))
// and any error encountered that caused the write to stop early.
// Write must return a non-nil error if it returns n < len(p).
// Write must not modify the slice data, even temporarily.
```

So I think it's fine ?

vickenty: Yes should be fine then. Good find.

statsd/uds_test.go
```diff
@@ -180,3 +183,39 @@ func TestUDSStreamWriteUnsetConnection(t *testing.T) {
...
+	err = os.Chmod(socketPath, 0722)
+	require.NoError(t, err)
+
+	w, err := newUDSWriter(socketPath, 100*time.Millisecond, 1000*time.Millisecond, "")
+	require.Nil(t, err)
+	require.NotNil(t, w)
+
+	// Force a connection
+	w.ensureConnection()
+	conn, err := listener.Accept()
```
vickenty, line 206:
```suggestion
	conn, err := listener.Accept()
	require.NoError(err)
```

statsd/uds_test.go
```diff
@@ -180,3 +183,39 @@ func TestUDSStreamWriteUnsetConnection(t *testing.T) {
...
+	defer conn.Close()
+
+	// Set a very low buffer size to force a partial write, but still enough to write the header
+	w.conn.(*net.UnixConn).SetWriteBuffer(1)
```
vickenty, line 210:
```suggestion
	require.NoError(w.conn.(*net.UnixConn).SetWriteBuffer(1))
```

vickenty (approved):

PR #300, opened 2024-01-08
[CONTINT-3864] Send both origin the container-id and the ENTITY_ID

AliDatadog:
We would like customers to be able to retrieve container tags even when `DD_ENTITY_ID` is set to false. The current behavior does not send the container-id if the entity id is set.

Note that the admission controller will set the container-id on Kubernetes.

For those who do not want container-tags, several alternatives are possible:
- Send the tag `dd.internal.card=none`
- Inject `DD_ORIGIN_DETECTION_ENABLED=false` to the application pod

See this doc for reference

### DIFF
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -454,21 +454,14 @@ func newWithWriter(w Transport, o *Options, writerName string) (*Client, error) 
 		errorHandler:          o.errorHandler,
 	}
 
-	hasEntityID := false
 	// Inject values of DD_* environment variables as global tags.
 	for _, mapping := range ddEnvTagsMapping {
 		if value := os.Getenv(mapping.envName); value != "" {
-			if mapping.envName == ddEntityID {
-				hasEntityID = true
-			}
 			c.tags = append(c.tags, fmt.Sprintf("%s:%s", mapping.tagName, value))
 		}
 	}
 
-	if !hasEntityID {
-		initContainerID(o.containerID, isOriginDetectionEnabled(o, hasEntityID), isHostCgroupNamespace())
-	}
-
+	initContainerID(o.containerID, isOriginDetectionEnabled(o), isHostCgroupNamespace())
 	isUDS := writerName == writerNameUDS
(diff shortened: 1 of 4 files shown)


statsd/statsd.go
```diff
@@ -888,13 +881,11 @@ func (c *Client) Close() error {
...
-// DD_ORIGIN_DETECTION_ENABLED is explicitly set to false.
-func isOriginDetectionEnabled(o *Options, hasEntityID bool) bool {
-	if !o.originDetection || hasEntityID || o.containerID != "" {
+// Disable origin detection only in one of the following cases:
+// - DD_ORIGIN_DETECTION_ENABLED is explicitly set to false
+// - o.originDetection is explicitly set to false, which is true by default
+func isOriginDetectionEnabled(o *Options) bool {
+	if !o.originDetection || o.containerID != "" {
 		// originDetection is explicitly disabled
 		// or DD_ENTITY_ID was found
```
wdhif, line 890: This comment should also be updated.

wdhif: Also here: https://github.com/DataDog/datadog-go/blob/master/statsd/options.go#L378

statsd/statsd.go
```diff
@@ -886,9 +886,6 @@ func (c *Client) Close() error {
 // - o.originDetection is explicitly set to false, which is true by default
 func isOriginDetectionEnabled(o *Options) bool {
 	if !o.originDetection || o.containerID != "" {
-		// originDetection is explicitly disabled
-		// or DD_ENTITY_ID was found
```
wdhif, line 899: nit: I would have left the rest of the comment but that's not a blocker.

wdhif (dismissed):

wdhif:
Links to https://github.com/DataDog/datadogpy/pull/828

gh123man (approved):
LGTM

gh123man (approved):

wdhif (approved):

PR #302, opened 2024-03-07
Make sure we retrieve the cgroup inode when the cid is not found

AliDatadog:
This PR fixes an issue where the cgroup inode when the container id is `""`. 

During testing of the last commit of this [PR](https://github.com/DataDog/datadog-go/pull/291), we found out that  the inode was sent and container tags was finally retrieved.

But actually, it was only because we had another bug in container id resolution that was fixed [here](https://github.com/DataDog/datadog-go/pull/298). We were overriding a faulty container-id. We shouldn't override the container-id (actually it's not problematic). But we should at least retrieve the container-id when it's not found.

The new test cases should cover everything


statsd/container_linux.go
```diff
@@ -184,23 +184,27 @@ func inodeForPath(path string) string {
...
 		}
-
-		if cgroupFallback {
-			isHostCgroupNs := isHostCgroupNamespace()
-			if isHostCgroupNs {
-				containerID = readContainerID(cgroupPath)
-				return
-			}
-			containerID = readMountinfo(selfMountInfoPath)
-			if containerID != "" {
```
AliDatadog, line 201: The bug is just this line. All the rest is about moving stuff around to add unit tests

vboulineau (approved):

rayz (approved):

PR #303, opened 2024-03-11
Support for using `statsd.Client` as an `otel.MeterProvider`

imle:
Looking to get some comments and guidance on this approach before getting to far into the weeds.

After a number of attempts, I settled on implementing the MeterProvider interface as a passthrough for otel metrics to the `statsd.Client`. This keeps much of the pipelineing system from otel out of the way of keeping things lined up with how this library already sends metrics.

Outstanding Work:
- [ ] observable metrics, the whole reason this needs to be implemented in the statsd library
- [ ] verify that it is acceptable to send a `float64` as a `Counter` value
- [ ] cleanup and improve readability and repeated work


carlosroman:
I wonder if this will break our compatibility with older versions of Golang. At the moment `datadog-go` compiles and works with versions all the way back to `go 1.12`.
This is an interesting approach to being able to reuse our Golang client to send OTel metrics. I'll discuss this internally to see what people think.

carlosroman:
Closing as this PR has gotten very stale.

PR #304, opened 2024-03-18
[CONTINT-3937] Fix origin detection when the cgroup namespace is shared but is not host cgroup ns

AliDatadog:
We noticed e2e test failures on Amazon Linux ECS Optimized OS. It runs with cgroupv1. The `kind` nodes are using docker and the pods use containerd. All the pods are running containers in the same cgroup namespace but this ns is the one allocated to the docker nodes, not the host cgroup namespace. Thus, we don't verify that the cid is found in /proc/self/cgroup whereas we should. We also don't send the cgroup inode.

### DIFF
--- a/statsd/container_linux.go
+++ b/statsd/container_linux.go
@@ -198,13 +198,16 @@ func readCIDOrInode(userProvidedID, cgroupPath, selfMountInfoPath, defaultCgroup
 	}
 
 	if cgroupFallback {
-		if isHostCgroupNs {
-			containerID = readContainerID(cgroupPath)
+		containerID = readContainerID(cgroupPath)
+		if containerID != "" {
 			return
 		}
+
 		containerID = readMountinfo(selfMountInfoPath)
-		if containerID == "" {
-			containerID = getCgroupInode(defaultCgroupMountPath, cgroupPath)
+		if containerID != "" || isHostCgroupNs {
+			return
 		}
+
+		containerID = getCgroupInode(defaultCgroupMountPath, cgroupPath)
 	}
 }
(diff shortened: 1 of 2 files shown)


statsd/container_linux.go
```diff
@@ -198,13 +198,16 @@ func readCIDOrInode(userProvidedID, cgroupPath, selfMountInfoPath, defaultCgroup
...
-			containerID = readContainerID(cgroupPath)
+		containerID = readContainerID(cgroupPath)
+		if containerID != "" {
 			return
 		}
+
 		containerID = readMountinfo(selfMountInfoPath)
-		if containerID == "" {
-			containerID = getCgroupInode(defaultCgroupMountPath, cgroupPath)
+		if containerID != "" || isHostCgroupNs {
```
vboulineau, line 207: What's the purpose of checking `isHostCgroupNs` here?

AliDatadog: I think we don't want to send the cgroup inode if we're in host cgroup namespace. Otherwise we'll send it even if the app is not in a container

vboulineau: I see, there isn't a reliable way to detect that. I think the approximation saying if `/proc/self/cgroup` does not find anything while being **in host cgroup namespace**, you can return.
There should be a dedicated code path and comment for it.

AliDatadog: Shouldn't an app that is running directly on host run in the host cgroup ns ? I am not sure I understood where exactly we should move this check but I added a dedicated `if` condition with a comment right after reading mountinfo.

vboulineau: Yes that's correct. You can move the check before `mountinfo` but it does not really matter.

carlosroman (approved):

PR #307, opened 2024-05-24
uds: add exponential backoff for reconnect

iksaif:


statsd/uds.go
```diff
@@ -96,11 +105,28 @@ func (w *udsWriter) Write(data []byte) (int, error) {
...
+	if stream {
+		// For uds-stream we want to gradually increase the write timeout
+		// up to the connect timeout to avoid overloading the agent with new connections.
+		// We increase by 20% or 500ms up to the connection timeout.
+		incr := w.currentWriteTimeout / 100 * 20
+		if incr == time.Duration(0) {
+			incr = time.Millisecond * 500
+		}
+		w.currentWriteTimeout += incr
+		if w.currentWriteTimeout > w.connectTimeout {
```
vickenty, line 124: This assumes that connection timeout is larger than the write timeout, although this might not be always the case. Would it make sense to ensure that connectTimeout is at least as big as writeTimeout, or perhaps use a separate upper bound?

PR #308, opened 2024-06-02
Optimistic metric allocation

opsengine:
Allocating new metrics before the lock is acquired for `count()` and `set()`. The `if` condition in the critical section is more likely to be false than true, so this will reduce the critical section and mostly without making useless allocations.

This optimization is already done on `gauge()` but for some reason is missing in the other two functions.

### DIFF
--- a/statsd/aggregator.go
+++ b/statsd/aggregator.go
@@ -219,6 +219,8 @@ func (a *aggregator) count(name string, value int64, tags []string) error {
 	}
 	a.countsM.RUnlock()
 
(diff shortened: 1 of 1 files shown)


remeh (approved):
Hey @opsengine,
Thank you for the contribution, it looks good I'm merging this one in!
I'll do some internals validation as well and then work on releasing it in a new version of the lib.

remeh:
Oh unfortunately, we can only merge signed commits. Any chance you could push force a signed commit instead?

PR #314, opened 2024-08-01
fix: Support cloning clients that use DD_AGENT_HOST

### LINKED ISSUE #313: client.addrOption remains unset when using default DD_AGENT_HOST
Reproduce steps 
```
// export DD_AGENT_HOST=whatever
client, err := statsd.New("", opts)
cloned, err := statsd.CloneWithExtraOptions(client, ...)
```
observe an error `can't clone client with no addrOption`

In the New function, [createWriter](https://github.com/DataDog/datadog-go/blob/master/statsd/statsd.go#L404) resolves the address in the function but it does not get set to the addOption in the client. It made the client not cloneable

abhinav:
Today, statsd client that use DD_AGENT_HOST for their address
cannot be cloned:

    // export DD_AGENT_HOST=localhost:8125
    client, err := statsd.New("")
    if err != nil {
      log.Fatal(err)
    }

    cloned, err := statsd.CloneWithExtraOptions(client, /* ... */)
    fmt.Println(err)  // can't clone client with no addrOption

This is because we resolve the address in createWriter,
but do not record the resolved address in the client.

    func New(addr string, options ...Option) (*Client, error) {
      // ...
      ... := createWriter(addr, /* ... */) // resolves addr internally
      // ...
      client.addrOption = addr  // addr is old, unresolved value
    }

This commit fixes this issue by replacing the address
with the resolved value so that that's what we record on the client,
allowing it to be cloned.

Testing: Includes a regression test for the bug.

Resolves #313

### DIFF
--- a/statsd/statsd_test.go
+++ b/statsd/statsd_test.go
@@ -206,6 +206,25 @@ func TestCloneWithExtraOptions(t *testing.T) {
 	assert.Len(t, cloneClient.options, 3)
 }
 
+func TestCloneWithExtraOptionsAddressFromEnvironment(t *testing.T) {
+	hostInitialValue, hostInitiallySet := os.LookupEnv(agentHostEnvVarName)
+	if hostInitiallySet {
+		defer os.Setenv(agentHostEnvVarName, hostInitialValue)
+	} else {
+		defer os.Unsetenv(agentHostEnvVarName)
+	}
+
+	_ = os.Setenv(agentHostEnvVarName, "localhost:1201")
+	client, err := New("", WithTags([]string{"tag1", "tag2"}))
+	require.NoError(t, err)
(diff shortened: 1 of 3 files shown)


abhinav (commented):
Note that newTelemetryClientWithCustomAddr is the only other caller of createWriter, so where previously it relied on createWriter to resolve the address, it must now do it before calling createWriter.

statsd/statsd_test.go
```diff
@@ -206,6 +206,25 @@ func TestCloneWithExtraOptions(t *testing.T) {
 	assert.Len(t, cloneClient.options, 3)
 }
 
+func TestCloneWithExtraOptionsAddressFromEnvironment(t *testing.T) {
```
carlosroman, line 209: Thank you for adding a test.

carlosroman (approved):

PR #326, opened 2025-07-01
[AGTMETRICS] send `DD_EXTERNAL_ENV` as `e:` field

brianna-wang-dd:


statsd/statsd.go
```diff
@@ -288,6 +288,8 @@ type Client struct {
 	isClosed              bool
 	errorOnBlockedChannel bool
 	errorHandler          ErrorHandler
+	// externalEnv is appended to every metric if there is a DD_EXTERNAL_ENV value
+	externalEnv string
```
StephenWakely, line 292:
```suggestion
```

The external env is stored in a global variable, so this field isn't being used and can be removed.

StephenWakely (changes requested):
Nicely done! Just a minor nit..

statsd/statsd.go
brianna-wang-dd: Thanks for the catch!

statsd/options.go
```diff
@@ -90,6 +90,11 @@ func resolveOptions(options []Option) (*Options, error) {
 	return o, nil
 }
 
+type ExtraOption interface{}
```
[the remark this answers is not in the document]
brianna-wang-dd: OOPS accidentally committed on the wrong branch. Just reverted!

rayz (approved):

PR #327, opened 2025-07-07
[AGTMETRICS-309] add support for tag cardinality

brianna-wang-dd:
This PR implements tag cardinality for this library by sending a cardinality field for each metric as `card:<cardinality>` field. Possible tag cardinality values are `none`, `low`, `orchestrator`, or `high`. 
`system.mangos:21|g#foo:bar,baz:boo|card:high`

The tag cardinality can be specified as a global setting (through the user passing the option to the constructor when setting up `Statsd` or by reading the value from `DD_CARDINALITY` or `DATADOG_CARDINALITY`). This can be overridden for each metric when creating one.

In order to ensure backward compatibility, a variadic parameters are added to each metric function that allow the tag cardinality to be passed in as type `CardinalityParameter`.

This PR was tested with unit test for tag cardinality specific functionality (such as global initialization from the 3 possible sources, checking the validity of inputs, and getting the global tag cardinality) and general metric functions (which included testing the metric-specific overrides for the tag cardinality).

### DIFF
--- a/statsd/tag_cardinality.go
+++ b/statsd/tag_cardinality.go
@@ -79,14 +79,14 @@ func getTagCardinality() Cardinality {
 }
 
 func parseTagCardinality(parameters []Parameter) Cardinality {
-	cardinality := getTagCardinality()
+	cardinality := CardinalityNotSet
 	for _, o := range parameters {
 		c, ok := o.(Cardinality)
 		if ok {
-			cardinality = resolveCardinality(c)
+			cardinality = c
 		}
 	}
-	return cardinality
+	return resolveCardinality(cardinality)
 }
 
 // resolveCardinality returns the cardinality to use, prioritizing the metric-level cardinality over the global setting.


rayz (changes requested):
If the benchmarks are showing that the adding the variadic parameters don't degrade performance, we can probably get rid of the `funky` metrics.

statsd/tag_cardinality.go
```diff
@@ -0,0 +1,57 @@
...
+
+	// If the user has not provided a valid value, read the value from the DD_CARDINALITY environment variable.
+	if tagCardinality.card == "" {
+		tagCardinality = validateCardinality(os.Getenv("DD_CARDINALITY"))
+	}
+	// If DD_CARDINALITY is not set or valid, read the value from the DATADOG_CARDINALITY environment variable.
+	if tagCardinality.card == "" {
+		tagCardinality = validateCardinality(os.Getenv("DATADOG_CARDINALITY"))
+	}
+	tagCardinalityMutex.Unlock()
```
rayz, line 37: suggestion: use a defer statement on line 25 instead 
`defer tagCardinalityMutex.Unlock()`

statsd/format.go
```diff
@@ -291,3 +291,19 @@ func appendExternalEnv(buffer []byte) []byte {
 	}
 	return buffer
 }
+
+func appendTagCardinality(buffer []byte, overrideCard CardinalityParameter) []byte {
+	// Check if the user has provided a valid cardinality parameter.
+	var value = validateCardinality(overrideCard.card).card
```
rayz, line 297:
```suggestion
	value := validateCardinality(overrideCard.card).card
```

statsd/options.go
```diff
@@ -412,3 +415,11 @@ func WithContainerID(id string) Option {
 		return nil
 	}
 }
+
+// WithCardinality sets the tag cardinality of the metric.
```
rayz, line 419: Could we update `options_test.go` too?

statsd/statsd.go
```diff
@@ -681,15 +685,23 @@ func (c *Client) sendToAggregator(mType metricType, name string, value float64,
...
 
 // Gauge measures the value of a metric at a particular time.
-func (c *Client) Gauge(name string, value float64, tags []string, rate float64) error {
+func (c *Client) Gauge(name string, value float64, tags []string, rate float64, parameters ...Parameter) error {
 	if c == nil {
 		return ErrNoClient
 	}
 	atomic.AddUint64(&c.telemetry.totalMetricsGauge, 1)
 	if c.agg != nil {
 		return c.agg.gauge(name, value, tags)
```
StephenWakely, line 694: A caveat I forgot to mention... 

The go client supports aggregating metrics. When aggregating, we need to take into account the cardinality, so metrics with different cardinality settings are not aggregated together. I think this will largely involve passing the cardinality through and including it as a part of the context in the GetContextAndTags. Then we need to make sure the cardinality is included when the metric gets flushed.

StephenWakely (changes requested):
Looking good! You can remove the Funky benchmarking code that you did since that has proved the code now. And then we need to ensure the cardinality is taken into consideration when aggregating.

statsd/statsd.go
```diff
@@ -681,15 +685,23 @@ func (c *Client) sendToAggregator(mType metricType, name string, value float64,
...
 	}
-	return c.send(metric{metricType: gauge, name: name, fvalue: value, tags: tags, rate: rate, globalTags: c.tags, namespace: c.namespace})
+
+	var cardinality = defaultTagCardinality
+	for _, o := range parameters {
+		c, ok := o.(CardinalityParameter)
+		if ok {
+			cardinality = c
+		}
+	}
```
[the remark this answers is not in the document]
brianna-wang-dd: Thanks for the catch!

statsd/aggregator.go
```diff
@@ -172,28 +172,38 @@ func (a *aggregator) flushMetrics() []metric {
...
 	}
+
 	n := len(name) + len(nameSeparatorSymbol) + len(tagSeparatorSymbol)*(len(tags)-1)
 	for _, s := range tags {
 		n += len(s)
 	}
 
+	var cardStringLen = 0
+	if cardinality.card != "" {
+		n += len("|card:") + len(cardinality.card)
```
StephenWakely, line 203: I think this can be slightly simplified. From the docs [here](https://docs.datadoghq.com/getting_started/tagging/#define-tags), we know that tags can't contain the `|` character, so instead of using `|card:` as the separator we can drop it to just `|`. It saves 5 bytes of memory, which would make it ever so slightly resource efficient, and it's just for internal use here so we don't need to worry about readability. It would be useful to define this as `cardSeparatorSymbol` as per `nameSeparatorSymbol`.

statsd/aggregator.go
```diff
@@ -204,13 +214,19 @@ func getContextAndTags(name string, tags []string) (string, string) {
...
 
+	if cardinality.card != "" {
+		sb.WriteString("|card:")
+		sb.WriteString(cardinality.card)
+	}
+
 	s := sb.String()
 
-	return s, s[len(name)+len(nameSeparatorSymbol):]
+	return s, s[len(name)+len(nameSeparatorSymbol) : len(s)-cardStringLen]
```
StephenWakely, line 224: Minute nit, but I think it would be slightly clearer if we had the cardinality field before the tags so the tags are always the last part of the context returned.

```suggestion
	return s, s[len(name)+len(nameSeparatorSymbol) + cardStringLen + len(cardSeparatorSymbol):]
```

statsd/aggregator.go
StephenWakely: I don't think this context is used when serialising any metrics, so it should be safe to change it - but if I'm wrong here obviously carry on as before!

statsd/aggregator_test.go
```diff
@@ -417,7 +431,461 @@ func BenchmarkGetContextNoTags(b *testing.B) {
 	name := "test.metric"
 	var tags []string
 	for i := 0; i < b.N; i++ {
-		getContext(name, tags)
+		getContext(name, tags, CardinalityParameter{card: "low"})
 	}
 	b.ReportAllocs()
 }
+
+func TestAggregatorCardinalitySeparation(t *testing.T) {
```
StephenWakely, line 439: Excellent test coverage!

statsd/tag_cardinality.go
```diff
@@ -0,0 +1,78 @@
...
+	"os"
+	"strings"
+	"sync"
+)
+
+type Parameter interface{}
+
+// Possible values for the cardinality are "none", "low", "orchestrator", and "high".
+type CardinalityParameter struct {
+	card string
```
StephenWakely, line 13: Given that cardinality is a limited number of options, really we should use more of an enum for this. Go doesn't have proper enums but I think something like this would make the intent clearer. Instead of doing string processing to handle the cardinality we just have integer operations, so slightly more efficient:

```go
type Cardinality int

const (
	None Cardinality = iota
	Low
	Orchestrator
	High
)

func (c Cardinality) String() string {
	switch c {
	case None:
		return "none"
	case Low:
		return "low"
	case Orchestrator:
		return "orchestrator"
	case High:
		return "high"
	}

	return ""
}
```

StephenWakely (changes requested):

statsd/aggregator.go
brianna-wang-dd: Yes, this makes sense especially since the context is just used internally to match metrics that should be aggregated together!

StephenWakely (dismissed):
Nicely done! I just have a few comments which are pretty pedantic so shouldn't block the PR, but feel free to address!

statsd/format.go
```diff
@@ -14,6 +14,7 @@ var (
 	timingSymbol        = []byte("ms")
 	tagSeparatorSymbol  = ","
 	nameSeparatorSymbol = ":"
+	cardSeparatorSymbol = "|"
```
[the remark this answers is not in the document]
brianna-wang-dd: Will do!

statsd/aggregator.go
```diff
@@ -206,11 +220,12 @@ func getContextAndTags(name string, tags []string) (string, string) {
...
 	s := sb.String()
 
-	return s, s[len(name)+len(nameSeparatorSymbol):]
+	return s, s[len(name)+len(nameSeparatorSymbol)+cardStringLen:]
 }
 
-func (a *aggregator) count(name string, value int64, tags []string) error {
-	context := getContext(name, tags)
+func (a *aggregator) count(name string, value int64, tags []string, cardinality Cardinality) error {
+	resolvedCardinality := resolveCardinality(cardinality)
```
AlexanderYastrebov: This is redundant

PR #330, opened 2025-07-17
[AGTMETRICS-319 ] Add check for origin detection configuration before outputting `e:` field

brianna-wang-dd:
This PR adds a check when appending the `e:` field from `DD_EXTERNAL_ENV` that sees if origin detection is configured to be on or not.  The `e:` field should not be appended if origin detection is configured off.


statsd/statsd.go
```diff
@@ -463,7 +463,8 @@ func newWithWriter(w Transport, o *Options, writerName string) (*Client, error)
 
 	initExternalEnv()
 
-	initContainerID(o.containerID, isOriginDetectionEnabled(o), isHostCgroupNamespace())
+	setOriginDetection(isOriginDetectionEnabled(o))
```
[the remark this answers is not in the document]
brianna-wang-dd: Thank you for the catch!! The origin detection should definitely be specific to each instance and not a global variable.

statsd/worker.go
```diff
@@ -108,21 +108,21 @@ func (w *worker) writeAggregatedMetricUnsafe(m metric, metricSymbol []byte, prec
 func (w *worker) writeMetricUnsafe(m metric) error {
 	switch m.metricType {
 	case gauge:
-		return w.buffer.writeGauge(m.namespace, m.globalTags, m.name, m.fvalue, m.tags, m.rate, m.timestamp)
+		return w.buffer.writeGauge(m.namespace, m.globalTags, m.name, m.fvalue, m.tags, m.rate, m.timestamp, m.originDetection)
```
vickenty, line 111: Would it make sense to store if originDetection is enabled inside the worker? Can it change after the client was created?

statsd/statsd.go
```diff
@@ -452,6 +454,7 @@ func newWithWriter(w Transport, o *Options, writerName string) (*Client, error)
 		telemetry:             &statsdTelemetry{},
 		errorOnBlockedChannel: o.channelModeErrorsWhenFull,
 		errorHandler:          o.errorHandler,
+		originDetection:       isOriginDetectionEnabled(o),
```
vickenty, line 457: `isOriginDetectionEnabled` returns false if containerID is set. Is it okay if we don't send external env in this case?

statsd/worker.go
brianna-wang-dd: I understand that the origin detection is set by passing in the options `WithOriginDetection()` or `WithoutOriginDetection()` or configuring through the environment variable, so I don't think that originDetection could change after the client was created. Looking at the `worker` struct, I see that it mainly holds the mechanisms for transporting the metrics. Would you recommend storing `originDetection` there instead of in the `client` struct and `metric` struct?

statsd/buffer.go
```diff
@@ -39,43 +39,44 @@ func newStatsdBuffer(maxSize, maxElements int) *statsdBuffer {
 	}
 }
 
-func (b *statsdBuffer) writeGauge(namespace string, globalTags []string, name string, value float64, tags []string, rate float64, timestamp int64, overrideCard Cardinality) error {
+func (b *statsdBuffer) writeGauge(namespace string, globalTags []string, name string, value float64, tags []string, rate float64, timestamp int64, originDetection bool, overrideCard Cardinality) error {
```
StephenWakely, line 42: I think that instead of passing `originDetection` through all these functions, it would be cleaner to just not call `initExternalEnv` [here](https://github.com/DataDog/datadog-go/blob/master/statsd/statsd.go#L465). This way the external env doesn't get populated and won't be written out. Is there a reason why we couldn't do it that way?

StephenWakely (changes requested):
I may have missed something, but I think there may be a cleaner way to avoid changing all the functions..

statsd/buffer.go
StephenWakely: Would it be possible to add a comment before calling `initExternalEnv` to make that clear so that someone doesn't come along later on and try to refactor it incorrectly.

brianna-wang-dd: Yep! I'll add a comment about the logic.

StephenWakely (approved):
Nice!

PR #336, opened 2025-09-12
pool string builders used in aggregator to concatenate tags

jstncbllr:
This reduced allocations significantly for us.

### DIFF
--- a/statsd/aggregator.go
+++ b/statsd/aggregator.go
@@ -181,6 +181,11 @@ func getContext(name string, tags []string) string {
 	return c
 }
 
+// stringBuilderPool pools strings.Builder objects to reduce allocations
+var stringBuilderPool = sync.Pool{
+	New: func() interface{} { return &strings.Builder{} },
+}
+
 // getContextAndTags returns the context and tags for a metric name and tags.
 //
 // See getContext for usage for context
@@ -194,7 +199,12 @@ func getContextAndTags(name string, tags []string) (string, string) {
 		n += len(s)
 	}
 
-	var sb strings.Builder
+	sb := stringBuilderPool.Get().(*strings.Builder)
+	defer func() {
+		sb.Reset()
+		stringBuilderPool.Put(sb)
+	}()
+
 	sb.Grow(n)
 	sb.WriteString(name)
 	sb.WriteString(nameSeparatorSymbol)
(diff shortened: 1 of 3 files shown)


carlosroman:
@review codex

statsd/aggregator.go
```diff
@@ -204,7 +209,12 @@ func getContextAndTags(name string, tags []string, cardinality Cardinality) (str
 		cardStringLen = len(cardString) + len(cardSeparatorSymbol)
 	}
 
-	var sb strings.Builder
+	sb := stringBuilderPool.Get().(*strings.Builder)
+	defer func() {
+		sb.Reset()
```
atanzu: I have a concern about correctness of this approach.

AFAIR `sb.String()` returns a string backed directly by the builder's internal buffer, so it does not copy. The defer here resets and returns the builder to the pool before the caller ever uses the returned strings, meaning the next goroutine to get the builder will overwrite the buffer that the returned string values still point into. This is silent heap corruption under concurrent use and might not be caught by the race detector.

atanzu (changes requested):

atanzu:
Hi @jstncbllr ,

Thank you for suggesting this improvement!
I left a comment because this part looks a bit suspicious to me. Could you please share your thoughts?

statsd/aggregator.go
jstncbllr: Thanks for digging into this. You were right to be suspicious, though the failure mode isn't quite there. `Builder.String()` does alias the internal buffer, but `Builder.Reset()` sets `buf = nil` rather than truncating, so the returned strings keep sole ownership of the old buffer and the next pool user allocates a fresh one. No corruption; a hammer test under the race detector agrees.

The real problem was one level up: because `Reset()` drops the buffer, the pool never reused any memory, so this PR didn't actually reduce allocations. Same allocs/op, slightly slower. A heap profile from one of our production services running this patch confirmed it was still allocating on every call. My original benchmark comparison was misleading since `getContextAndTagsOld` predated the cardinality changes and the two benchmarks weren't building the same string.

I pushed a rework earlier today that builds the key into a pooled byte buffer and probes the map with `m[string(buf)]`, which avoids the allocation on the lookup path. Then I noticed #389 already landed a more complete version of the same idea on master: stack buffers instead of a pool, hashing while appending for shard selection, and a no-tags fast path. So there's nothing left for this PR to do and I'm closing it. Thanks for the review.

PR #354, opened 2026-02-10
Upgrade go-winio from v0.5.0 to v0.6.2 (CVE-2025-65637)

nadavshatz:
go-winio v0.5.0 depended on sirupsen/logrus v1.7.0, which is vulnerable to CVE-2025-65637 — a high-severity (CVSS 7.5) denial of service where Entry.Writer() crashes on single-line payloads exceeding 64KB due to bufio.Scanner's token size limit, rendering the writer permanently unusable.

The latest v0.5.x release (v0.5.3) still depends on logrus v1.7.0 and remains vulnerable, so a patch-level upgrade within the v0.5.x line is not sufficient to remediate this CVE. Upgrading to v0.6.2 eliminates the vulnerable logrus transitive dependency from the build graph entirely.

The minimum Go version is bumped from 1.13 to 1.21 as required by the updated dependency chain.


nadavshatz:
Any chance to get this reviewed?

carlosroman:
Sorry, I’m not able to accept this change currently because it introduces incompatibilities with several versions of Go that we still support. I will start  a discussion internally to see if we are ready to deprecate support for these older versions of Go. Dropping support for older Go versions will require a major version bump for the library. If we are going to do that we might want to introduce some other breaking changes.

nadavshatz:
I completely understand. Please also discuss a solution for the dependency with the CVE?

nadavshatz:
Thanks for the context @carlosroman, totally understand the concern about dropping older Go versions.

Would you be open to a smaller PR that addresses the CVE without bumping the Go minimum version? Two options that keep go-winio at v0.5.x:

**Option A — `replace` directive:**
```go
replace github.com/sirupsen/logrus v1.7.0 => github.com/sirupsen/logrus v1.9.3
```

**Option B — `exclude` directive:**
```go
exclude github.com/sirupsen/logrus v1.7.0
```

Both would force the module graph to resolve logrus to a non-vulnerable version while keeping go-winio at v0.5.0 and the Go minimum version unchanged.

Happy to open a new PR with either approach if that works for you.

PR #358, opened 2026-02-24
Replace vulnerable logrus v1.7.0 transitive dependency (CVE-2025-65637)

nadavshatz:
## Summary

- Adds a `replace` directive in `go.mod` to redirect the vulnerable `sirupsen/logrus v1.7.0` (transitive dependency of `go-winio v0.5.0`) to `v1.9.3`
- Remediates CVE-2025-65637 (CVSS 7.5 DoS via `Entry.Writer()` on payloads >64KB) without bumping the Go minimum version or upgrading go-winio
- No changes to library code or public API — only `go.mod` and `go.sum` are affected

This is a lighter alternative to #354, which was declined because upgrading go-winio to v0.6.2 required bumping the minimum Go version from 1.13 to 1.21. See [discussion](https://github.com/DataDog/datadog-go/pull/354#issuecomment-3950375200).

## Test plan

- [ ] `go mod tidy` succeeds without errors
- [ ] `go mod graph | grep logrus` confirms logrus resolves to v1.9.3
- [ ] Existing CI tests pass with no changes


carlosroman:
Having a look at why the tests are failing under Go 1.12. It should just pass.

nadavshatz:
@carlosroman I found the root cause and pushed a fix.

**Problem:** The `replace` directive redirects logrus to v1.9.3, which declares `testify v1.7.0` in its `go.mod`. The `go_12_test` job already uses `sed` to downgrade testify to v1.6.0, but Go 1.12 lacks [lazy module loading](https://go.dev/ref/mod#lazy-loading) (introduced in Go 1.17), so MVS still resolves `testify v1.7.0` from the logrus transitive dependency — and that version uses `errors.Is`/`errors.As` which require Go 1.13+.

**Fix:** Added a second `sed` step in `go_12_test` to strip `replace` directives from `go.mod` before running the tests. This is consistent with the existing testify downgrade workaround, and since the module already declares `go 1.13` as its minimum, the replace directive isn't needed for the Go 1.12 best-effort compat test.

nadavshatz:
Hi maintainers — friendly bump on this security-related PR. The follow-up commit addressed the Go 1.12 CI issue mentioned earlier, checks are now green, and the branch is mergeable. Would appreciate a review when possible.

carlosroman:
@nadavshatz thanks, sorry I missed your update. Will gives this another try.

atanzu (approved):

atanzu:
Hi @nadavshatz ,

Thank you a lot for your contribution!

nadavshatz:
thanks for merging! looking forward to a new version to be released.

PR #359, opened 2026-02-27
Replace vulnerable logrus v1.7.0 transitive dependency (CVE-2025-65637) with v1.8.3

carlosroman:


atanzu:
Superseeded in #386 .

PR #379, opened 2026-05-18
test: make tests independent of `containerID` ordering

KowalskiThomas:
## What is this PR?

While trying to make some changes on `datadog-go`, I've noticed that some tests would not work when run all at the same time. This should help with that when running

````
go test ./statsd/ -timeout 300s -race -count=1
````

Specifically, the container ID detection logic (in `container.go` / `container_linux.go`) could return multiple candidate IDs, and tests were asserting on a specific one. The fix refactors the container detection to expose a deterministic/sorted result and updates the test helpers and test files (`container_test.go`, `statsd_test.go`, `telemetry_test.go`, `service_check_test.go`) to not rely on a particular ordering of container IDs.

### CODE REVIEW

File: statsd/format_test.go [hunk 1]
```diff
@@ -44,84 +46,98 @@ func TestFormatAppendTagsAggregated(t *testing.T) {
...
 
 func TestFormatAppendHistogram(t *testing.T) {
+	withoutOriginGlobals(t)
 	var buffer []byte
 	buffer = appendHistogram(buffer, "namespace.", []string{"global:tag"}, "histogram", 3., []string{"tag:tag"}, 1, true)
 	assert.Equal(t, `namespace.histogram:3|h|#global:tag,tag:tag`, string(buffer))
 }
 
 func TestFormatAppendDistribution(t *testing.T) {
+	withoutOriginGlobals(t)
```
vickenty, line 70: Last parameter of appendGauge controls whether the container ID will be added. Would it be sufficient to set it to false?

statsd/telemetry_test.go
```diff
@@ -18,13 +18,19 @@ import (
 //
 
 func TestTelemetryCustomAddr(t *testing.T) {
+	// The expected strings below were written for an empty container ID;
+	// disable origin detection so the test is hermetic and does not depend
+	// on /proc/self/cgroup or on test ordering populating the package
+	// global.
+	withoutOriginGlobals(t)
```
vickenty, line 25: Even if we do have a container ID, `WithoutOriginDetection` alone should prevent the client from appending it to the metric. If this is not true, the tests have found an issue in the production code.

KowalskiThomas: To both your comments, `originDetection` only controls whether `appendExternalEnv` does anything (if it's `false` then it just returns), `appendContainerID` on the other hand always appends the container ID. I don't know if you Container ID is considered as external env?

carlosroman:
Strange, race detector has picked up an issue with some of the tests on Windows. Might be a platform specific race condition (never seen one) or could be specific to Windows code we have.

statsd/test_helpers_test.go
```diff
@@ -644,13 +644,27 @@ func (ts *testServer) sendExtendedBasicAggregationMetricsWithPreAggregatedSample
...
 
+// withoutOriginGlobals clears the package-level containerID and externalEnv
+// before the test runs and restores them after, so the test is independent
+// of any other test in the package that may have populated them through a
+// real client construction. Tests that build clients and assert wire
+// payloads which do not include container-id / external-env tags should
+// call this at the very top of the test body.
+func withoutOriginGlobals(t *testing.T) {
+	t.Helper()
+	resetContainerID()
```
atanzu, line 663: I think the way we use `withoutOriginGlobals` is the reason for the detected race condition, because we are now resetting `containerID` without any synchronization. From the trace we see `TestServiceChecks` calls `withoutOriginGlobals()` while a telemetry goroutine from an earlier client is concurrently formatting telemetry and reading `containerID` in `appendContainerID()`.

statsd/test_helpers_test.go
```diff
@@ -644,13 +644,27 @@ func (ts *testServer) sendExtendedBasicAggregationMetricsWithPreAggregatedSample
...
+	containerID = id
+}
 
 func resetContainerID() {
 	containerID = ""
 	initOnce = sync.Once{}
 }
 
+// withoutOriginGlobals clears the package-level containerID and externalEnv
+// before the test runs and restores them after, so the test is independent
```
atanzu: Where do we restore those?

atanzu (changes requested):

statsd/format_test.go
KowalskiThomas: _(as noted in my response on your other comment: `appendContainerID` on the other hand always appends the container ID regardless of the value of `originDetection`)_

statsd/test_helpers_test.go
KowalskiThomas: Yep, I think that's correct. I'm trying to make `containerID` an atomic like I previously did for `externalEnv` in the other PR, let's see if it helps.

statsd/test_helpers_test.go
KowalskiThomas: That's a leftover comment from some fixing back-and-forth; I think we don't currently restore them but we should. IIRC the problem is that `t.Cleanup` wasn't available in all required Go versions. Anyway, I'll keep that specific thing for later once I've fixed the race issue. I'll let you know!

KowalskiThomas: Okay so here's what I did
- I updated the comment to remove the reference to `Cleanup` since we don't have it
- I updated every test that relies on origin globals being unset / empty to call `withoutOriginGlobals` so test ordering doesn't matter 

To be honest I don't think the "clear before relevant tests" approach is necessarily worse than the "clean up after relevant tests", especially since we don't have a proper way to clean up, so I'm personally fine with the current version.

Let me know what you think.

atanzu (approved):

PR #381, opened 2026-05-19
perf: flip condition order

KowalskiThomas:
## What is this PR?

Micro optimisation -- the function is about 2.5x faster when `originDetection` is `false` compared to previously.


vickenty (approved):

KowalskiThomas:
@carlosroman I believe Vickenty told me the other day the Merge Queue doesn't work on datadog-go

PR #389, opened 2026-06-12
Improve client performance

atanzu:
This PR introduces a series of changes aimed to improve performance of the Go client.

# Overview

| Area | Improvement | Trade-off |
|---|---|---|
| Flush (any shard count) | −78–82% time, −97–100% allocs | None |
| Construction (high shard count) | −38% time, −99% allocs | Memory +154% at 256 shards (one-time, ~68 KB) |
| Sharding (low shard count ≤4) | −45–58% time, −100% allocs | None |
| Sharding (high shard count ≥64) | −18–24% time, −100% allocs | None |
| Steady-state sampling (Tags ≤60) | −34–46% time, 3→0 allocs | None |
| Steady-state sampling (Tags_100) | −5% time | 3 allocs unchanged (large tag sets still escape) |
| Per-metric-type sampling | −29–40% time, 1→0 allocs | None |

---

# More detailed overview

---

## Change 0 - New benchmark

Because we need more numbers to judge the changes.

---

## Change 1 - New shard allocation and placement logic

Improve aggregator's construction and flush. Previously each shard allocated its own map on construction and its own slice on every flush cycle. Both are now amortised/eliminated. Steady-state sampling is unchanged.

**Construction — allocs/op (deterministic)**

| Shards | Before | After | Delta |
|---|---|---|---|
| 1 | 24 | 18 | −25% |
| 32 | 210 | 18 | −91% |
| 256 | 1,554 | 18 | −99% |

**Construction — B/op and sec/op**

| Shards | Before B | After B | Before t | After t |
|---|---|---|---|---|
| 1 | 16.9 KB | 16.8 KB | 23.4 µs | 23.4 µs |
| 32 | 24.9 KB | 20.0 KB | 26.1 µs | 23.6 µs |
| 256 | 83.4 KB | 44.4 KB | 45.1 µs | 25.0 µs |

**Flush — allocs/op (deterministic)**

| Benchmark | Before | After | Delta |
|---|---|---|---|
| FlushEmpty/Shards_1 | 6 | 3 | −50% |
| FlushEmpty/Shards_32 | 99 | 3 | −97% |
| FlushEmpty/Shards_256 | 771 | 3 | −99% |
| FlushSparse | 789 | 24 | −97% |

**Flush — sec/op and B/op**

| Benchmark | Before t | After t | Before B | After B |
|---|---|---|---|---|
| FlushEmpty/Shards_1 | 152 ns | 88 ns | 288 B | 144 B |
| FlushEmpty/Shards_32 | 2,494 ns | 505 ns | 4,752 B | 144 B |
| FlushEmpty/Shards_256 | 19,643 ns | 3,552 ns | 37,008 B | 144 B |
| FlushSparse | 21,335 ns | 4,587 ns | 39.1 KB | 3.2 KB |

**Geomean (sec/op):** −5.2% · **Geomean (allocs/op):** −18.6%

---

## Change 2 - Optimize metric contexts builds

Improves the steady-state sampling hot path a.k.a. the code executed every time a metric is recorded against an existing context. Previously this allocated on every call for most tag counts; after this commit it is zero-allocation for Tags ≤ ~60.

**AggregatorSampleExisting — allocs/op (deterministic)**

All shard counts, Tags_5 through Tags_60: **3 allocs → 0**
Tags_100: unchanged at 3 allocs (large tag sets still escape to heap)

**AggregatorSampleExistingByMetricType — allocs/op**

Count / Gauge / Set: **1 alloc → 0**

**BufferedMetricContextsSampleExisting — allocs/op**

All shapes below Tags_100: **1 alloc → 0** (Tags_100: 1 → 0 as well)

**AggregatorSampleExisting — sec/op (representative: Shards_8)**

| Tags | Before | After | Delta |
|---|---|---|---|
| 5 | 195 ns | 112 ns | −42% |
| 10 | 314 ns | 185 ns | −41% |
| 40 | 932 ns | 569 ns | −39% |
| 60 | 1,277 ns | 823 ns | −36% |
| 100 | 1,974 ns | 1,843 ns | −7% |

**AggregatorSampleExisting — sec/op (Shards_16, showing tag-count scaling)**

| Tags | Before | After | Delta |
|---|---|---|---|
| 5 | 192 ns | 102 ns | −47% |
| 10 | 322 ns | 176 ns | −45% |
| 40 | 945 ns | 564 ns | −40% |
| 60 | 1,245 ns | 819 ns | −34% |

**AggregatorSampleExistingByMetricType — sec/op**

| Metric type | Before | After | Delta |
|---|---|---|---|
| Count | 15.6 ns | 10.9 ns | −30% |
| Gauge | 15.4 ns | 10.7 ns | −30% |
| Set | 17.7 ns | 13.7 ns | −23% |

**BufferedMetricContextsSampleExistingByShape — sec/op**

| Shape | Before | After | Delta |
|---|---|---|---|
| NoTags_Low | 242 ns | 230 ns | −5% |
| FiveTags_Low | 270 ns | 243 ns | −10% |
| Tags100_Low | 430 ns | 342 ns | −20% |
| Tags100_High | 429 ns | 338 ns | −21% |

**Geomean (sec/op):** −20.2% 

---

## Change 3 - Pad aggregator shards to avoid false sharing

Use cache-line padding on shard structs to eliminate false sharing between goroutines on different CPUs accessing different shards. 

**AggregatorSharding — sec/op**

| Shards | Before | After | Delta |
|---|---|---|---|
| 1 | 431 ns | 281 ns | −35% |
| 2 | 327 ns | 139 ns | −58% |
| 3 | 213 ns | 98 ns | −54% |
| 4 | 157 ns | 79 ns | −49% |
| 8 | 91 ns | 62 ns | −32% |
| 16 | 63 ns | 49 ns | −21% |
| 32 | 54 ns | 44 ns | −18% |
| 64 | 46 ns | 43 ns | −7% |
| 128 | 43 ns | 40 ns | −8% |
| 256 | 41 ns | 40 ns | −4% |

**AggregatorSampleExisting at low shard counts — sec/op**

False-sharing removal shows through in sampling too, especially at Shards_1–4:

| Config | Before | After | Delta |
|---|---|---|---|
| Shards_1 / Tags_5 | 382 ns | 130 ns | −66% |
| Shards_1 / Tags_40 | 397 ns | 154 ns | −61% |
| Shards_2 / Tags_5 | 241 ns | 119 ns | −51% |
| Shards_4 / Tags_5 | 141 ns | 104 ns | −26% |
| Shards_8 / Tags_5 | 112 ns | 103 ns | −8% |
| Shards_16+ / Tags_5 | ~100 ns | ~98 ns | ≈ flat |

**AggregatorSampleExistingByMetricType — sec/op**

| Metric type | Before | After | Delta |
|---|---|---|---|
| Count | 10.9 ns | 9.0 ns | −17% |
| Gauge | 10.7 ns | 9.0 ns | −16% |
| Set | 13.7 ns | 11.2 ns | −18% |

**Geomean (sec/op):** −10.2%

### Trade-offs / regressions

**Downside:** Padding inflates the in-memory size of each shard struct, so construction memory grows proportionally to shard count. 

| Shards | Before B | After B | Delta |
|---|---|---|---|
| 1 | 16.8 KB | 17.0 KB | +2% |
| 32 | 20.0 KB | 30.9 KB | +54% |
| 256 | 44.4 KB | 112.7 KB | +154% |

This memory is allocated once at client construction and held for the lifetime of the
client, so the impact is negligible for long-running processes. At 256 shards the
footprint grows by ~68 KB.

---

# Benchmark environment:
- Apple M4 Max, 16 cores
- go1.26.3 darwin/arm64
- canary spread 1.01×
- 3 interleaved rounds

### DIFF
--- a/.github/workflows/datadog-go.yaml
+++ b/.github/workflows/datadog-go.yaml
@@ -35,8 +35,13 @@ jobs:
             go-version: 1.14
           - runs-on: macos-latest
             go-version: 1.15
+          # These fail on macOS 26+ because of this bug https://github.com/golang/go/issues/68678
           - runs-on: macos-latest
             go-version: 1.21
+          - runs-on: macos-latest
+            go-version: 1.22
+          - runs-on: macos-latest
+            go-version: 1.23
           # No official linux/arm64 Go binaries before 1.16
           - runs-on: ubuntu-24.04-arm
             go-version: 1.13
--- a/statsd/aggregator.go
+++ b/statsd/aggregator.go
@@ -35,9 +35,9 @@ type aggregator struct {
 	nbContextSet   uint64
 
 	shardsCount int
-	countShards []*countShard
-	gaugeShards []*gaugeShard
-	setShards   []*setShard
+	countShards []countShard
+	gaugeShards []gaugeShard
+	setShards   []setShard
 
 	histograms    bufferedMetricContexts
 	distributions bufferedMetricContexts
@@ -60,19 +60,14 @@ func newAggregator(c *ClientEx, maxSamplesPerContext int64, shardsCount int) *ag
 	agg := &aggregator{
 		client:          c,
 		shardsCount:     shardsCount,
-		countShards:     make([]*countShard, shardsCount),
-		gaugeShards:     make([]*gaugeShard, shardsCount),
-		setShards:       make([]*setShard, shardsCount),
+		countShards:     make([]countShard, shardsCount),
+		gaugeShards:     make([]gaugeShard, shardsCount),
+		setShards:       make([]setShard, shardsCount),
 		histograms:      newBufferedContexts(newHistogramMetric, maxSamplesPerContext),
 		distributions:   newBufferedContexts(newDistributionMetric, maxSamplesPerContext),
 		timings:         newBufferedContexts(newTimingMetric, maxSamplesPerContext),
 		closed:          make(chan struct{}),
 		stopChannelMode: make(chan struct{}),
-	}
-	for i := 0; i < shardsCount; i++ {
-		agg.countShards[i] = &countShard{counts: countsMap{}}
-		agg.gaugeShards[i] = &gaugeShard{gauges: gaugesMap{}}
-		agg.setShards[i] = &setShard{sets: setsMap{}}
 	}
 	return agg
 }
@@ -155,10 +150,15 @@ func (a *aggregator) flushMetrics() []metric {
 	// We reset the values to avoid sending 'zero' values for metrics not
 	// sampled during this flush interval
 
-	for _, shard := range a.setShards {
+	for i := range a.setShards {
+		shard := &a.setShards[i]
 		shard.Lock()
 		sets := shard.sets
-		shard.sets = setsMap{}
+		if len(sets) == 0 {
+			shard.Unlock()
+			continue
+		}
+		shard.sets = nil
 		shard.Unlock()
 		for _, s := range sets {
 			metrics = append(metrics, s.flushUnsafe()...)
@@ -166,10 +166,15 @@ func (a *aggregator) flushMetrics() []metric {
 		atomic.AddUint64(&a.nbContextSet, uint64(len(sets)))
 	}
 
-	for _, shard := range a.gaugeShards {
+	for i := range a.gaugeShards {
+		shard := &a.gaugeShards[i]
 		shard.Lock()
 		gauges := shard.gauges
-		shard.gauges = gaugesMap{}
+		if len(gauges) == 0 {
+			shard.Unlock()
+			continue
+		}
+		shard.gauges = nil
 		shard.Unlock()
 		for _, g := range gauges {
 			metrics = append(metrics, g.flushUnsafe())
@@ -177,10 +182,15 @@ func (a *aggregator) flushMetrics() []metric {
 		atomic.AddUint64(&a.nbContextGauge, uint64(len(gauges)))
 	}
 
-	for _, shard := range a.countShards {
+	for i := range a.countShards {
+		shard := &a.countShards[i]
 		shard.Lock()
 		counts := shard.counts
-		shard.counts = countsMap{}
+		if len(counts) == 0 {
+			shard.Unlock()
+			continue
+		}
+		shard.counts = nil
 		shard.Unlock()
 		for _, c := range counts {
 			metrics = append(metrics, c.flushUnsafe())
@@ -255,7 +265,7 @@ func getShardIndex(shardsCount int, context string) int {
 
 func (a *aggregator) count(name string, value int64, tags []string, cardinality Cardinality) error {
 	context := getContext(name, tags, cardinality)
-	shard := a.countShards[getShardIndex(a.shardsCount, context)]
+	shard := &a.countShards[getShardIndex(a.shardsCount, context)]
 	shard.RLock()
 	if count, found := shard.counts[context]; found {
 		count.sample(value)
@@ -274,6 +284,9 @@ func (a *aggregator) count(name string, value int64, tags []string, cardinality 
 		return nil
 	}
 
+	if shard.counts == nil {
+		shard.counts = countsMap{}
+	}
 	shard.counts[context] = metric
 	shard.Unlock()
 	return nil
@@ -281,7 +294,7 @@ func (a *aggregator) count(name string, value int64, tags []string, cardinality 
 
 func (a *aggregator) gauge(name string, value float64, tags []string, cardinality Cardinality) error {
 	context := getContext(name, tags, cardinality)
-	shard := a.gaugeShards[getShardIndex(a.shardsCount, context)]
+	shard := &a.gaugeShards[getShardIndex(a.shardsCount, context)]
 	shard.RLock()
 	if gauge, found := shard.gauges[context]; found {
 		gauge.sample(value)
@@ -299,6 +312,9 @@ func (a *aggregator) gauge(name string, value float64, tags []string, cardinalit
 		shard.Unlock()
 		return nil
 	}
+	if shard.gauges == nil {
+		shard.gauges = gaugesMap{}
+	}
 	shard.gauges[context] = gauge
 	shard.Unlock()
 	return nil
@@ -306,7 +322,7 @@ func (a *aggregator) gauge(name string, value float64, tags []string, cardinalit
 
 func (a *aggregator) set(name string, value string, tags []string, cardinality Cardinality) error {
 	context := getContext(name, tags, cardinality)
-	shard := a.setShards[getShardIndex(a.shardsCount, context)]
+	shard := &a.setShards[getShardIndex(a.shardsCount, context)]
 	shard.RLock()
 	if set, found := shard.sets[context]; found {
 		set.sample(value)
@@ -323,6 +339,9 @@ func (a *aggregator) set(name string, value string, tags []string, cardinality C
 		set.sample(value)
 		shard.Unlock()
 		return nil
+	}
+	if shard.sets == nil {
+		shard.sets = setsMap{}
 	}
 	shard.sets[context] = metric
 	shard.Unlock()
--- a/statsd/aggregator.go
+++ b/statsd/aggregator.go
@@ -14,6 +14,11 @@ type (
 	bufferedMetricMap map[string]*bufferedMetric
 )
 
+const (
+	smallContextBufferSize = 512
+	largeContextBufferSize = 4 * 1024
+)
+
 type countShard struct {
 	sync.RWMutex
 	counts countsMap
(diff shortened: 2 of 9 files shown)


statsd/aggregator.go
```diff
@@ -14,19 +15,33 @@ type (
 	bufferedMetricMap map[string]*bufferedMetric
 )
 
+const (
+	// cacheLineSize is the size shards are padded to so that adjacent shards never
+	// share a cache line. Apple silicon and some other arm64 cores use 128-byte
+	// lines, so we pad to 128 to cover every platform we run on. Without this,
+	// several shards pack into one line and an RLock on one shard (which writes the
+	// mutex's reader count) bounces the neighbouring shards' mutexes between cores.
+	cacheLineSize          = 128
```
tobz, line 23: Nit: might be cleaner to use [CacheLinePad](https://cs.opensource.google/go/x/sys/+/v0.46.0:cpu/cpu.go;l=22) [¶](https://pkg.go.dev/golang.org/x/sys/cpu#CacheLinePad).

statsd/aggregator.go
```diff
@@ -14,19 +15,33 @@ type (
...
+	smallContextBufferSize = 512
+	largeContextBufferSize = 4 * 1024
+)
+
 type countShard struct {
 	sync.RWMutex
 	counts countsMap
+	_      [cacheLineSize - (unsafe.Sizeof(sync.RWMutex{})+unsafe.Sizeof(countsMap(nil)))%cacheLineSize]byte
```
tobz: I wonder if you could define these structs in an inner/outer form, like this:

```go
type countShardInner struct {
   sync.RWMutex
   counts countsMap
}

type countShard struct {
    countShardInner
    _ [cacheLineSize - unsafe.Sizeof(countShard{}) % cacheLineSize]byte
}
```

Might be entirely valid Go, but just a thought on defining things in parts so the padding math is a little more durable to change.

statsd/aggregator.go
atanzu: Many thanks! That's the bit I was missing to keep it elegant.

statsd/aggregator.go
atanzu: Good catch, done.

statsd/aggregator.go
atanzu: This won't work that nice here. The `cacheLineSize` from the mentioned package is not exported, and prepending the whole thing won't give the same effect in terms of the overall size.

statsd/aggregator.go
```diff
@@ -274,14 +451,78 @@ func (a *aggregator) count(name string, value int64, tags []string, cardinality
...
+}
+
+// gaugeWithLargeContextBuffer keeps the 4 KiB stack array out of gauge's frame
+// so the common small-context path is not penalized by the larger frame.
+func (a *aggregator) gaugeWithLargeContextBuffer(contextLen int, name string, value float64, tags []string, cardinality Cardinality, cardString string) error {
+	if contextLen <= largeContextBufferSize {
+		var contextBuffer [largeContextBufferSize]byte
+		return a.gaugeWithContextBuffer(contextBuffer[:0], name, value, tags, cardinality, cardString)
+	}
+	return a.gaugeWithContextBuffer(make([]byte, 0, contextLen), name, value, tags, cardinality, cardString)
```
vickenty, line 434: Would it make sense to add some tests to make sure we correctly aggregating metrics with different context sizes?

statsd/aggregator.go
```diff
@@ -246,16 +288,151 @@ func getContextAndTags(name string, tags []string, cardinality Cardinality) (str
...
+}
+
+func (a *aggregator) countWithContextBuffer(contextBuffer []byte, name string, value int64, tags []string, cardinality Cardinality, cardString string) error {
+	contextHash := uint32(0)
+	if a.shardsCount > 1 {
+		contextBuffer, contextHash, _ = appendContextAndHash(contextBuffer, name, tags, cardString)
+	} else {
+		contextBuffer, _ = appendContext(contextBuffer, name, tags, cardString)
+	}
+	shard := &a.countShards[getShardIndexFromHash(a.shardsCount, contextHash)]
```
vickenty, line 352: Code above this line is the same for all flavors. Would we loose on performance if we extracted this repeated bit into a method?

Same question for per-shard code below, which is repeated in countWithStringContext.

statsd/aggregator.go
```diff
@@ -246,16 +288,151 @@ func getContextAndTags(name string, tags []string, cardinality Cardinality) (str
...
+	for _, tag := range tags {
+		n += len(tag)
+	}
+	if cardString != "" {
+		n += len(cardString) + len(cardSeparatorSymbol)
+	}
+	return n
+}
+
+func appendContextAndHash(buf []byte, name string, tags []string, cardString string) ([]byte, uint32, int) {
```
vickenty: Am I missing something, or is the last return value is never used?

statsd/aggregator.go
```diff
@@ -246,16 +288,151 @@ func getContextAndTags(name string, tags []string, cardinality Cardinality) (str
...
+	if shard.counts == nil {
+		shard.counts = countsMap{}
+	}
+	context := string(contextBuffer)
+	shard.counts[context] = metric
+	shard.Unlock()
+	return nil
+}
+
+func (a *aggregator) countWithStringContext(context string, contextHash uint32, name string, value int64, tags []string, cardinality Cardinality) error {
```
vickenty: Since this method assumes tags and cardinality are not set, would it make sense to remove these arguments, so we can never pass wrong values?

statsd/buffered_metric_context.go
```diff
@@ -89,12 +142,8 @@ func (bc *bufferedMetricContexts) sample(name string, value float64, tags []stri
...
 	}
 
-	// Now we can keep the sample or skip it
-	if keepingSample {
-		v.maybeKeepSample(value, bc.random, &bc.randomLock)
-	} else {
-		v.skipSample()
-	}
+	// Now we can keep the sample.
+	v.maybeKeepSample(value, bc.random, &bc.randomLock)
```
vickenty, line 146: skipSample is no longer called, is it intentional?

statsd/buffered_metric_context.go
```diff
@@ -55,32 +55,85 @@ func (bc *bufferedMetricContexts) flush(metrics []metric) []metric {
...
+}
+
+// sampleWithLargeContextBuffer keeps the 4 KiB stack array out of sample's
+// frame so the common small-context path is not penalized by the larger frame.
+func (bc *bufferedMetricContexts) sampleWithLargeContextBuffer(contextLen int, name string, value float64, tags []string, rate float64, cardinality Cardinality, cardString string) error {
+	if contextLen <= largeContextBufferSize {
+		var contextBuffer [largeContextBufferSize]byte
+		return bc.sampleWithContextBuffer(contextBuffer[:0], name, value, tags, rate, cardinality, cardString)
+	}
+	return bc.sampleWithContextBuffer(make([]byte, 0, contextLen), name, value, tags, rate, cardinality, cardString)
```
vickenty, line 114: Would it make sense to add tests to make sure we correctly go through all these branches?

statsd/aggregator.go
atanzu: I don't think so. In the very end we append to a slice; once we exceed the capacity, Go runtime allocates a new backing array on the heap.

statsd/aggregator.go
atanzu: Unfortunately yes, I tried to compact those common pieces into shared functions and got hit by a performance penalty. If I understand the problem correctly, Go compiler is limited in optimizing deep nested function calls.

statsd/aggregator.go
atanzu: Yes, this is an artifact from the optimization iterations, will remove.

statsd/buffered_metric_context.go
atanzu: Good catch — `skipSample` is dead code now that the early return in `sample()` handles rate-filtering. I've removed it. The TODO comment already documents the plan to revisit rate tracking later when we switch to proper per-thread structures.

statsd/buffered_metric_context.go
atanzu: The three paths differ only in *how* the buffer is allocated — the resulting context key (and all downstream logic) is identical regardless of which branch is taken. The existing tests already verify context key correctness through the sampler; adding size-specific tests would require synthetic metrics with carefully calibrated name+tag lengths, which feels brittle for what is purely an allocation strategy rather than a logic branch.

statsd/aggregator.go
atanzu: Yes, makes sense. That's another generational artifact.

statsd/buffered_metric_context.go
vickenty: > which feels brittle for what is purely an allocation strategy rather than a logic branch.

I think it is still worth while. Right now one can pass `nil` for `tags` in the large case  and the tests wouldn't catch it.

atanzu: I still don't quite get the concern. AFAIR there is no limitation on the metric's  name length (at least on this side), so if a user sends a metric with a name of 3000 symbols and no tags, that's still a valid case which needs the largest buffer available.

statsd/aggregator.go
```diff
@@ -246,18 +288,122 @@ func getContextAndTags(name string, tags []string, cardinality Cardinality) (str
...
 
 func (a *aggregator) count(name string, value int64, tags []string, cardinality Cardinality) error {
-	context := getContext(name, tags, cardinality)
-	shard := a.countShards[getShardIndex(a.shardsCount, context)]
+	if len(tags) == 0 && cardinality == CardinalityNotSet {
+		contextHash := uint32(0)
+		if a.shardsCount > 1 {
+			contextHash = hashString32(name)
+		}
+		return a.countWithStringContext(name, contextHash, name, value)
```
vickenty, line 323: It seems that we don't have any tests without tags. Since this is a new code  path, would it make sense to add some?

vickenty (commented):
Are getContext and getContextAndTags dead code now?

statsd/aggregator.go
atanzu: `getContext` and `getContextAndTags` were needed till mid-branch and now indeed those are dead code. I've pushed another commit to replace the obsolete tests with the ones verifying the actual code.

statsd/buffered_metric_context.go
vickenty: The concern is that _we_ might pass nil instead of user tags by mistake in, for example, huge allocating case, and the tests will not catch that. It is easy to see that the code is identical between three context buffers today, but we should still test it.

atanzu: Makes sense now, thanks. I've added a regression test to check against that.

statsd/aggregator.go
```diff
@@ -195,69 +236,122 @@ func (a *aggregator) flushMetrics() []metric {
...
-func getContextAndTags(name string, tags []string, cardinality Cardinality) (string, string) {
-	cardString := cardinality.String()
+func getContextLength(name string, tags []string, cardString string) int {
 	if len(tags) == 0 {
 		if cardString == "" {
-			return name, ""
+			return len(name)
 		}
-		return name + nameSeparatorSymbol + cardString, ""
+		return len(name) + len(nameSeparatorSymbol) + len(cardString)
```
vickenty, line 244: These branches (and matching ones in `appendContext` and `appendContextAndHash`) seem to be lacking test coverage. Would it be possible to add it?

.github/workflows/datadog-go.yaml
```diff
@@ -37,6 +37,10 @@ jobs:
             go-version: 1.15
           - runs-on: macos-latest
             go-version: 1.21
+          - runs-on: macos-latest
```
vickenty, line 41: Is it the same bug linked above? If not, would it be possible to link to the bug report?

atanzu: No, that's a different one.
I've added https://github.com/golang/go/issues/68678 to the comments.

vickenty (approved):

PR #395, opened 2026-07-28
Fix reservoir sampling off-by-one error

atanzu:
Increment totalSamples first and draw from the post-increment count instead to match Algorithm R and fix the double-duty of totalSamples being read both atomically (`LoadInt64`) and non-atomically (`s.totalSamples++`) under the same lock.

`maybeKeepSample()` drew its replacement index from `s.totalSamples` before incrementing it for the current sample, so once the reservoir was full it drew from `[0, k-1)` instead of `[0, k)` for the k-th observed sample. Algorithm R requires the current sample to be counted in that range: replacement probability should be `maxSamples / k`, not `maxSamples / (k-1)`.

Also since we only access `totalSamples` under a lock, we remove all the atomic reads of it to reduce the performance penalty brought by atomic ops.

### DIFF
--- a/statsd/metrics.go
+++ b/statsd/metrics.go
@@ -174,11 +174,12 @@ func (s *bufferedMetric) maybeKeepSample(v float64, rand *rand.Rand, randLock *s
 	s.Lock()
 	defer s.Unlock()
 	if s.maxSamples > 0 {
+		total := atomic.AddInt64(&s.totalSamples, 1)
 		if s.storedSamples >= s.maxSamples {
 			// We reached the maximum number of samples we can keep in memory, so we randomly
 			// replace a sample.
 			randLock.Lock()
-			i := rand.Int63n(atomic.LoadInt64(&s.totalSamples))
+			i := rand.Int63n(total)
 			randLock.Unlock()
 			if i < s.maxSamples {
 				s.data[i] = v
@@ -187,7 +188,6 @@ func (s *bufferedMetric) maybeKeepSample(v float64, rand *rand.Rand, randLock *s
 			s.data[s.storedSamples] = v
 			s.storedSamples++
 		}
-		s.totalSamples++
 	} else {
 		// This code path appends to the slice since we did not pre-allocate memory in this case.
 		s.sampleUnsafe(v)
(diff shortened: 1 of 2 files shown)


statsd/metrics.go
```diff
@@ -174,11 +174,12 @@ func (s *bufferedMetric) maybeKeepSample(v float64, rand *rand.Rand, randLock *s
 	s.Lock()
 	defer s.Unlock()
 	if s.maxSamples > 0 {
+		total := atomic.AddInt64(&s.totalSamples, 1)
```
vickenty: Is there any reason to use an atomic operation here since s.Lock is always held?

atanzu: Actually no, and what is even more interesting: I've grepped the codebase and found out that we always access it under a lock, so this and other places can be updated to omit the atomic ops, which I've done in the follow-up commit.

statsd/metrics.go
```diff
@@ -166,19 +166,20 @@ func (s *bufferedMetric) sample(v float64) {
...
 	defer s.Unlock()
 	if s.maxSamples > 0 {
+		s.totalSamples++
+		total := s.totalSamples
 		if s.storedSamples >= s.maxSamples {
 			// We reached the maximum number of samples we can keep in memory, so we randomly
 			// replace a sample.
 			randLock.Lock()
-			i := rand.Int63n(atomic.LoadInt64(&s.totalSamples))
+			i := rand.Int63n(total)
```
vickenty:
```suggestion
			i := rand.Int63n(s.totalSamples)
```

statsd/metrics.go
```diff
@@ -166,19 +166,20 @@ func (s *bufferedMetric) sample(v float64) {
...
-	atomic.AddInt64(&s.totalSamples, 1)
+	s.totalSamples++
 }
 
 func (s *bufferedMetric) maybeKeepSample(v float64, rand *rand.Rand, randLock *sync.Mutex) {
 	s.Lock()
 	defer s.Unlock()
 	if s.maxSamples > 0 {
+		s.totalSamples++
+		total := s.totalSamples
```
vickenty:
```suggestion
```

vickenty (dismissed):

vickenty (approved):

PR #399, opened 2026-08-18
Support IPv6 DogStatsD addresses

### LINKED ISSUE #275: Client Support IPv6 address
The current `New` function supports taking an IPv4 address, but not an IPv6 address for initialization i.e.

```go
func TestIPv6(t *testing.T) {
	// passes
	_, err := statsd.New("127.0.0.1")
	assert.NoError(t, err)

	// fails
	_, err = statsd.New("::1")
	assert.NoError(t, err)
}
```

It would be _really nice_ if the `New` function would support IPv6 addresses in the same way to make service migrations to IPv6 not require code changes : )

BakedSoups:
Hey, big fan of the Go migration here! I saw #275 and wanted to give a PR a shot.

# Problem

The address resolver assumed that any address containing a colon already included a port. Raw IPv6 addresses such as `::1` therefore reached the UDP writer without brackets or a port.

# Solution

Use Go’s standard address helpers to normalize IPv6 addresses while preserving explicit ports. This supports IPv6 addresses passed directly, through `DD_AGENT_HOST`, or through `DD_DOGSTATSD_URL`. Zone identifiers are supported for direct addresses and `DD_AGENT_HOST`.

# Testing

- `go test -race ./statsd -run ^TestResolveAddressFromEnvironment$ -count=1`
- `go vet ./...`

Fixes #275.

### DIFF
--- a/statsd/statsdex.go
+++ b/statsd/statsdex.go
@@ -16,6 +16,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/url"
 	"os"
 	"strconv"
@@ -344,16 +345,20 @@ func resolveAddr(addr string) string {
 			return addr
 		}
 	}
-	// TODO: How does this work for IPv6?
-	if strings.Contains(addr, ":") {
+	if envPort != "" {
+		return withPort(addr, envPort)
+	}
+	return withPort(addr, defaultUDPPort)
+}
+
+func withPort(addr, port string) string {
+	if _, _, err := net.SplitHostPort(addr); err == nil {
 		return addr
 	}
-	if envPort != "" {
-		addr = fmt.Sprintf("%s:%s", addr, envPort)
-	} else {
-		addr = fmt.Sprintf("%s:%s", addr, defaultUDPPort)
-	}
-	return addr
+
+	host := strings.TrimPrefix(addr, "[")
+	host = strings.TrimSuffix(host, "]")
+	return net.JoinHostPort(host, port)
 }
 
 func parseAgentURL(agentURL string) string {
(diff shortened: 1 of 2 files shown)


statsd/statsdex.go
```diff
@@ -344,16 +345,26 @@ func resolveAddr(addr string) string {
...
 		return addr
 	}
-	if envPort != "" {
-		addr = fmt.Sprintf("%s:%s", addr, envPort)
-	} else {
-		addr = fmt.Sprintf("%s:%s", addr, defaultUDPPort)
+
+	// Preserve the existing handling of host:port-like addresses while treating
+	// addresses with multiple colons as unbracketed IPv6 literals.
+	if strings.Count(addr, ":") == 1 {
```
atanzu: AFAIR `SplitHostPort` already handles things like `localhost:`, `:1234`, `foo:bar` (basically any input with a single `:`) so this check might probably be unneeded.

atanzu:
The PR description claims zone identifier support, and it works for direct addresses and `DD_AGENT_HOST` like e.g. `DD_AGENT_HOST=fe80::1%lo0`, but zone IDs do not work in the `DD_DOGSTATSD_URL` path because url.Parse interprets % as a URL escape sequence, so `udp://fe80::1%lo0` or `udp://[fe80::1%eth0]` will fail. Could you please adjust the PR description to clarify that zone ID support works with `DD_AGENT_HOST` and direct addresses, not `DD_DOGSTATSD_URL`?

BakedSoups:
Thanks for the review! I removed the redundant single-colon check since `net.SplitHostPort` already handles those cases.

I also updated the PR description to clarify that zone identifiers are supported for direct addresses and `DD_AGENT_HOST`, but not through `DD_DOGSTATSD_URL`.

PR #403, opened 2026-08-25
Drop golang.org/x/net

atanzu:
Drop dependency on `golang.org/x/net` which we use only for local testing. Replace it with a helper which returns a temporary Unix socket path.


go.mod
```diff
@@ -6,7 +6,10 @@ require (
 	github.com/Microsoft/go-winio v0.5.0
 	github.com/golang/mock v1.6.0
 	github.com/stretchr/testify v1.8.1
-	golang.org/x/net v0.0.0-20210405180319-a5a99cb37ef4
 )
 
 replace github.com/sirupsen/logrus v1.7.0 => github.com/sirupsen/logrus v1.9.3
+
+replace golang.org/x/crypto => golang.org/x/crypto v0.17.0
```
vickenty, line 14: Would it make sense to leave a note why these replace directives are needed?

vickenty (dismissed):

go.mod
atanzu: Fair, amended the commit.

vickenty (approved):

